PR #8, opened 2021-04-02
Fixed asd saving bug!

hasrat-ali:


JavkhlanK:
YES! It works! How did you even find out it should be an Object?

hasrat-ali:
I just took a deep look at LogicEditorActivity.

PR #11, opened 2021-04-05
fix: Assets and Native Libraries Path

hasrat-ali:


JavkhlanK:
Oh my, how did I forget to check the path? 😅

PR #12, opened 2021-04-05
fix: Intent keys and some menus

hasrat-ali:
Intent keys have been replaced with old ones.

Some menu patterns were updated for better selection.


JavkhlanK:
Yeah, it's better to display most used intent extras, and if the users needs different ones, they can use Code Editor's input instead. Good change👍

PR #13, opened 2021-04-07
feat : Removed 'Activity' string from 'Fragments'.

hasrat-ali:
This file is not fully optimised.


JavkhlanK:
What do you mean by "not fully optimised"?

hasrat-ali:
This means that it is not readable like your files.

There is no bug in it.

JavkhlanK:
Alright, I'll clean it up later.

PR #14, opened 2021-04-08
Create SketchDialog.java

hasrat-ali:


JavkhlanK:
Any use of it until now? Also, having public fields and setters, but no getters isn't good. Can you either make (most) fields private and add necessary getters or remove the setters?

hasrat-ali:
So far there is no use, but it has been used in VersionDialog.

JavkhlanK:
now some getters are missing, like `getDialogImage()`

hasrat-ali:
What is the use of getDialogImage().

JavkhlanK:
It should return `dialog_img`, and also be public.

hasrat-ali:
I have updated but I do not think it is necessary.

JavkhlanK:
Those getters would be great for all private Views in that class.

PR #30, opened 2021-04-12
feat: Fully implemented IncrementalJavaCompiler

tyron12233:
Compilation of java files is fully working and is being generated on `/mysc/incremental/classes` however it's not yet integrated with Dp.java


app/src/main/java/mod/tyron/compiler/IncrementalJavaCompiler.java
```diff
@@ -235,8 +352,110 @@ public CompilerOutputStream(StringBuffer buffer) {
...
+            n.a("gson-2.8.0");
+        }
+        if (projectConfig.N.n) {
+            n.a("glide-4.11.0");
+        }
+        if (projectConfig.N.p) {
+            n.a("okhttp-3.9.1");
+        }
+        ExtLibSelected.a(projectConfig.N.x, n);
+    }
```
JavkhlanK, line 460: These methods seem to be copied from Dp.java. Could you use them instead, as you have a Dp object anyway?

app/src/main/java/mod/tyron/compiler/Util.java
```diff
@@ -0,0 +1,7 @@
+package mod.tyron.compiler;
+
+import java.io.File;
+
+public class Util {
+
+}
```
JavkhlanK, line 7: What's this class for?

JavkhlanK (changes requested):
There are 2 files that should be changed first. The rest of the files seem good, though.

app/src/main/java/mod/tyron/compiler/Util.java
tyron12233: Unused class, it will be used in the future ill remove it on this commit

tyron12233 (commented):
Removed dp methods

JavkhlanK (approved):
Seems good now :+1:

PR #36, opened 2021-04-13
add : dummy a.a.a to make the new pl compile

AlucardTN:


JavkhlanK:
The new what? Also, these files were anonymous classes that got obfuscated outside their package name. Please decompile these files, and put them as anonymous classes in the decompiled class that references them.

MarkDroid:
add those because yeb, they are some a.a.a fields generated to make some classes in com.besome.sketch.beans and editor compile !!!

Iyxan23:
hi bor

DantaNotEpic:
real conversation here, adios

MarkDroid:
i cant decmpile all those a.a.a classes man!, just they are a o onclick listener, and which they are not even needed

JavkhlanK:
mmm alright I'll add them now, but then decompile and bring their functionality back to the original uses later

MarkDroid:
yes, okay, this may take many time

AlucardTN:
merge !!!

PR #41, opened 2021-04-13
Changed about activity

ilyassesalama:
For some reason, I changed android manifest too.


AlucardTN:
ilyasse, you don't need to push .idea files, add a gitignore + you need to implement the new about activity with material 1.0.0 because sketchware uses and we can't change it because sketchware classes uses ( it's obfuscated too )

ilyassesalama:
About activity uses material 1.0.0

AlucardTN:
oh, i didnt notice because you must remove app compat and mateial from your build.gradle

ilyassesalama:
I left them for testing purposes. Only RecyclerView requires androidx in the whole activity.

AlucardTN:
im not sure but i think recyclerview has low chance to work, because its obfuscated but we will give it a try dont worry

AlucardTN:
+ dont forget the gitignore :horny:

ilyassesalama:
Hmm. In that case, I'll close this pull request and switch to ListView instead.

AlucardTN:
its merged now, let firstly jbk0 test it then we will notify you if worked or not, the pl is already merged

PR #70, opened 2021-04-18
feat : added Version Dialog option

hasrat-ali:


hasrat-ali:
I have tested and all works fine.

JavkhlanK:
waiting for the new dialog being implemented now 
apensivefast

Issue #73, opened 2021-04-19
Sketchware Pro can't read previously added fragment's events


hasrat-ali:
Previously added fragment's events cannot be read due to new `getActivityName`.

JavkhlanK:
Fixed with #74.

PR #74, opened 2021-04-19
fix : issue #73 not reading fragment's events

hasrat-ali:


hasrat-ali:
If we add a new fragment, it performs well but all the events of the previously added fragment disappear.

It cannot read any of the events of the previously added fragment.

So I have removed this feature.

JavkhlanK:
What did the old feature even do? I haven't really looked at its code before.

hasrat-ali:
It's generate `ChatFragment` string instead of `ChatFragmentActivity` of fragment.

PR #75, opened 2021-04-19
feat : add some a.a.a classes !!

AlucardTN:


JavkhlanK:
I'm sorry, but many files inside a.a.a are obfuscated anonymous classes (because their only usage is inside a class from the package com.besome.sketch), and they shouldn't be decompiled like this, but rather inlined into the only classes referencing them.
Also, I've been working on most com.besome.sketch classes lately, and wanted to test the changes out soon. Sorry that you wasted a lot of time on nothing useful.

PR #76, opened 2021-04-19
feat, fix, style : see commits. 

AlucardTN:
DesignActivity is high file even i used CRLF and design want to detect new lines or changes, btw i only changed things of compile from line proguard


AlucardTN:
also, hosni forget to rename hes methods, so i fixed them

AlucardTN:
thats all, still some bugs about incremental build

PR #80, opened 2021-04-20
feat : add add_file_48 to Resources

AlucardTN:


JavkhlanK:
Looks like you (sorta) figured out how to make these, good job! It's just that the resource ID type apparently is not a drawable. All other drawable IDs start with `0x7f070`, but your added constant starts with `0x7f080`, which means it's a view ID (normally something similar to `R.id.linear1`)

Please make sure to put the constant into the right position inside the class, meaning constants with a higher value get defined later inside the classes. (Check if the hex value fits in, the fields should be sorted)

PR #81, opened 2021-04-20
feat : add LayoutUtilities to make copy of sketchware toolbar

AlucardTN:


JavkhlanK:
Alright, but did you see the `layou` in the package name? Also, not lowercase methods aren't really Java-like (`LayoutUtilities#Dp(Context, int)`). Another thing to note on that method is that `mod.SketchwareUtil` already provides that.

And also, the class uses `com.jbk.swproremodremod.R`. That means the resource IDs **will** be invalid (later) in Sketchware Pro. Please use `com.sketchware.remod.Resources` instead.

AlucardTN:
1 button can resolve everything

PR #91, opened 2021-04-25
fix : unusual Custom Variable Spec

hasrat-ali:


JavkhlanK:
Has this caused any issues before?

hasrat-ali:
No

PR #95, opened 2021-04-26
feat: added uq.java

hasrat-ali:
Added some extra Intent Actions.


AlucardTN:
did you tested the changes out ?

hasrat-ali:
Yes

AlucardTN:
so it should be accepted, thanks !

PR #96, opened 2021-04-26
feat: added MB.java

hasrat-ali:
This file is for Validators.

tested out.


AlucardTN:
if the file is tested and working, so you need to push it into directly into branch main yourself, **this only for new files**

PR #130, opened 2021-05-17
del: remove R8Executor

AlucardTN:


JavkhlanK:
Dp.java does have a reference to it, so it's not useless right now. It's just not finished.

AlucardTN:
it's not implemented there, means i can remove it you are using if (false) means the method will not even executed son, remove that

JavkhlanK:
why remove it tho?

AlucardTN:
because

DantaNotEpic:
big por modir moment

AlucardTN:
shut

DantaNotEpic:
rud <a:fastsad:816072729148260353>

PR #131, opened 2021-05-17
del: remove LayoutUtilities

AlucardTN:
because useless


JavkhlanK:
well, this class *does* have no references, so...

PR #133, opened 2021-05-18
feat, improvement: some classes

hasrat-ali:


JavkhlanK:
Alright, looks good 👍
Could you try to use 4 spaces the next time tho please? Most of the files on the repo have them too, so the changes won't be too different.

PR #136, opened 2021-06-02
feat: added code editor menu

hasrat-ali:


JavkhlanK:
I don't think having `static` `Context`s is good.

hasrat-ali:
i don't understand you

tyron12233 (commented):
Why is everything static? It has a constructor and it doesnt make sense to make it static

hasrat-ali:
ok, i will resolve it.

PR #138, opened 2021-06-05
Some cleanups and optimizations

Iyxan23:
Some cleanups and optimizations on `mod/agus` and `mod/alucard`

Note: I haven't tested this, but looking from the changes, there shouldn't be any difference


JavkhlanK:
🙏

PR #156, opened 2021-07-02
feat, fix: added missing blocks, fixes list block

hasrat-ali:
`feat:` added `SignInButton`, `OTPView`, and `PageAdapter`'s blocks and menu.

`fix:` fixes list's blocks not showing


JavkhlanK:
I thought we're not going to add more built-in components?

hasrat-ali:
> I thought we're not going to add more built-in components?

Yes me too, but this is the last.

PR #165, opened 2021-07-13
Feat: Added feature to clean project trash

RealNethical:
just Added a Project trash cleaner . It deletes all trash files inside the [projectnum]/bin folder except the classes folder as deleting it will decrease the compilation time .


JavkhlanK:
note: Discussed in Discord server.

PR #166, opened 2021-07-14
Feat: Added feature to clean project trash

RealNethical:
just Added a Project trash cleaner . It deletes all trash files inside the [projectnum]/bin folder except the classes folder as deleting it will decrease the compilation time .


JavkhlanK:
Alright, these commits are messy, but I'll fix them.
Is it okay if I'll rename some methods, fields, etc.?
And also, the cleaner would be an option inside the previously Build Settings dialog button's PopupMenu then.

RealNethical:
> Alright, these commits are messy, but I'll fix them.
> Is it okay if I'll rename some methods, fields, etc.?
> And also, the cleaner would be an option inside the previously Build Settings dialog button's PopupMenu then.

Yep bro
U can do anything .

PR #167, opened 2021-07-15
feat: set a role to the subtitle

hosni-fraj:
Now subtitle will show the project id


JavkhlanK:
not that bad, but isn't the sc_id in MainActivity's project list enough?

hosni-fraj:
But we need to add this as a feature, some of user can't remember it, they will used to go back then he will waste his time as.

JavkhlanK:
fair enough, approved ✅

PR #168, opened 2021-07-15
feat: added some comments to the DesignActivity

hosni-fraj:
feat: added some comments


hosni-fraj:
ofc comments for the new code

JavkhlanK:
I don't think it's needed, the code above clearly has `"sc_id"` in it, and the Toolbar method wasn't obfuscated.

PR #174, opened 2021-07-17
Feat : added new feature to rename libraries 

RealNethical:


app/src/main/java/dev/aldi/sayuti/editor/manage/ManageLocalLibraryActivity.java
```diff
@@ -207,21 +208,43 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
...
-                        public boolean onMenuItemClick(MenuItem menuItem) {
-                            FileUtil.deleteFile(local_libs_path.concat(checkBox.getText().toString()));
-                            bB.a(ManageLocalLibraryActivity.this, "Deleted successfully", 0).show();
-                            loadFiles();
-                            return true;
-                        }
-                    });
-                    popupMenu.show();
+                    @Override
+    public boolean onMenuItemClick(MenuItem item){
```
JavkhlanK, line 212: Seems like a good idea, but this section is broken. You probably forgot to add something?

RealNethical: Actually its because of the text editor. We can fix later after commiting

PR #175, opened 2021-07-17
Added decompiled dt.class

Keller-Zirus:
For dt.java


app/src/main/java/a/a/a/dt.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    public void initRefresh() {
+        this.radio_mb_type.setOnCheckedChangeListener(new a(this));
+        radiobutton1.setOnCheckedChangeListener(new OnCheckedChangeListener() {
+			@Override
+			public void onCheckedChanged(RadioGroup radioGroup, int i) {
+				refresh(getType());
+			}
+		});
```
JavkhlanK, line 243: Seems like these lines wouldn't compile, could you check them again?

Keller-Zirus: I updated it.Can you try again?

PR #179, opened 2021-07-19
Update library_downloader_dialog.xml

BlueCatSoftware:
Changed edittext to textinputlayout

### DIFF
--- a/app/src/main/res/layout/library_downloader_dialog.xml
+++ b/app/src/main/res/layout/library_downloader_dialog.xml
@@ -51,7 +51,7 @@
             android:indeterminate="false" />
     </LinearLayout>
 
-    <LinearLayout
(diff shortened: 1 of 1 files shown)


JavkhlanK:
Should work fine, but that needs some changes in the Java file using it. Could you modify it too?

BlueCatSoftware:
> Should work fine, but that needs some changes in the Java file using it. Could you modify it too?

Yes, no Problem

BlueCatSoftware:
> Should work fine, but that needs some changes in the Java file using it. Could you modify it too?
What's the java file path? 
It should be in hey studio path

JavkhlanK:
It's used in `mod.hey.studios.project.library.LibraryDownloader`.

PR #185, opened 2021-07-20
MoreBlock with custom type

Keller-Zirus:
Can create moreblock with custom type

### DIFF
--- a/app/src/main/java/a/a/a/dt.java
+++ b/app/src/main/java/a/a/a/dt.java
@@ -1,0 +1,269 @@
+package a.a.a;
+
+import android.app.Activity;
+import android.graphics.Rect;
+import android.util.Pair;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.Button;
+import android.widget.EditText;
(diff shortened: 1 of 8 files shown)


JavkhlanK:
nice massive commits but what exactly happened to res/values/ids.xml? it's been modified a lot, but why? and also, your commit messages sadly don't follow the guidelines, please read them.

Keller-Zirus:
You mean,I don't need to modify res/values/ids.xml and rewrite commit massages.

Keller-Zirus:
> nice massive commits but what exactly happened to res/values/ids.xml? it's been modified a lot, but why? and also, your commit messages sadly don't follow the guidelines, please read them.

You mean,I don't need to modify res/values/ids.xml and rewrite commit massage.

JavkhlanK:
of course you can change ids.xml, but I was asking why there were 3k changes. if possible, please try to not add resource IDs as they're painful to manage and are probably way more effort than to not give views IDs. instead, you can create views at runtime and set their LayoutParams.
and also yeah, the commit messages should be re-written to follow the guidelines.

PR #186, opened 2021-07-20
feat: added custom type for Moreblock

Keller-Zirus:
Can create moreblock with custom moreblock type.
Hope this work.


Keller-Zirus (commented):
Aa

JavkhlanK:
rip but the commit messages are still not following the guidelines, but that's not important for now

PR #189, opened 2021-07-24
Feat: added logcat for easy debugging

Insight-deviler:
Now using the code `Log.i(TAG, "your message");` has a purpose in Sketchware project.
**Usage:**
1. Create a shared preference component with component name and filename as `log`.
2. To enable Logger add this code in your activity `log.edit().putString("debug", "true").commit();`
3. To Disable Logger add this `log.edit().putString("debug", "false").commit();`

> Important
> 1. Use this only for debugging and not when app is public as this may consume storage if unchecked.
> 2. The log files are stored in `/Android/data/{your application package}/log_time.txt`
> 3. Restart your application after enabling or disabling of logcat.
> 4. If you want to disable logger you need to either add disable code or clear app data as this is a shared preference

### DIFF
--- a/app/src/main/assets/debug/SketchApplication.java
+++ b/app/src/main/assets/debug/SketchApplication.java
@@ -22,7 +22,7 @@ import java.io.IOException;
 
 import android.widget.Toast;
 
-public class SketchApplication extends Application {
+public class InsightDeviler extends Application {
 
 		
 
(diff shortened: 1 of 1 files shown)


hasrat-ali:
Too messy code (bad formatted).

Insight-deviler:
Created in mobile phone that's why it's like that 😬. Will fix it soon

JavkhlanK:
Why are there that many newlines? lol

Insight-deviler:
I updated the class using mobile which resulted in many unwanted lines now I removed those lines

hasrat-ali:
Use spaces (`    `) instead of tabs `\t` as indentation charactor.

Insight-deviler (commented):
Removed those unnecessary spaces

tyron12233:
I suggest to write it on the app's external files directory instead of writing it in shared preferences

Insight-deviler:
It will be writing log data to data/your application package/log.
I used shared preference to enable or disable logging

khaled-0:
Add everything to build settings as a checkbox when.
 Or how can I contribute in a PR🤔

JavkhlanK:
This *might*'ve been a good idea back then when SketchApplication.java wasn't easily modify/replaceable, but now, I think we should rather provide this as a template, maybe in the Discord server, or the app?
Users could then apply such changes themselves, instead of having things like these in the stock SketchApplication.java.

PR #193, opened 2021-07-25
fix: Android 11 Scoped Storage. style: Removed Tutorials Tab

khaled-0:
**fix: Android 11 Scoped Storage Fixes.**
Adding request for 'ALL FILE ACCESS' and thus read and write no longer depends on emulated SAF on Android 11 or further.
Results: No more freezing or infinite loading time when saving or building a project.

P.S: Tested on my Device running android 11. And almost all the slowdowns are gone. It performs like older android 10 again

**style: Removed Tutorial tab from MainActivity**
Removed the not working unused tutorial tab from MainActivity .
Now it only contains 'MY PROJECTS'

Note : I just commented out certain parts of the PageAdapter so it won't include the tutorials tab. The fragment is still in there.


khaled-0:
> where are this method's imports lol

Inline import supremacy

PR #195, opened 2021-07-26
Fix: fixed crash in Local Library Manager

Keller-Zirus:
Fixed ConcurrentModificationException error


khaled-0:
Bruh, they both does the same thing.

Keller-Zirus:
> Bruh, they both does the same thing.

IMG_20210726_182048

khaled-0:
I would suggest using an Iterator so the code would look a bit nice.
But anyways if it works , it works 🐸

JavkhlanK:
I think the entire logic of the activity should be changed, or at least made efficient. For now though, this will be enough.

PR #201, opened 2021-07-28
removed Tablayout from main activity and added scrollview

Pranav-sketchware:
added scrollview in project configuration dialog (requested by Mike Anderson)
commented code eferencing to Tablayout and removed Tablayout setup with viewpager
removed Tablayout from xml


JavkhlanK:
Wouldn't removing the TabLayout in that way also remove the projects list? Also, does the new ScrollView need to have a resource ID?

Pranav-sketchware:
projectlist won't be gone, as I removed the viewpager setup

PR #202, opened 2021-07-28
style : Added a New look to the Debug Activity

Rakshit2:


hasrat-ali:
There is already exists `RequestNetwork` and `RequestNetworkController` in https://github.com/Sketchware-Pro/Sketchware-Pro/tree/main/app/src/main/java/mod

hasrat-ali:
Is there any uses of this.
https://github.com/Sketchware-Pro/Sketchware-Pro/blob/7f335116cf85544ae880ed71cc47063acf6abd2f/app/src/main/java/com/besome/sketch/Debug4Activity.java#L148-L213

Rakshit2:
> There is already exists `RequestNetwork` and `RequestNetworkController` in https://github.com/Sketchware-Pro/Sketchware-Pro/tree/main/app/src/main/java/mod

i didnt noticed sorry for that

Rakshit2:
i had to use an api

Rakshit2:
I am really sorry i forgot to put the extra string error

Rakshit2:
> Is there any uses of this.
> https://github.com/Sketchware-Pro/Sketchware-Pro/blob/7f335116cf85544ae880ed71cc47063acf6abd2f/app/src/main/java/com/besome/sketch/Debug4Activity.java#L148-L213

All are Waste except the on back pressed void

JavkhlanK:
Seems like a great idea! I'd rather move it into the existing DebugActivity though, but I'll do that later.

hasrat-ali (commented):
And where is `layout/debug4.xml`

Rakshit2:
Added the Layout XML file

hasrat-ali (commented):
Why are there minus characters in your `layout/debug4.xml` file.

Rakshit2:
> Why are there minus characters in your `layout/debug4.xml` file.

Internet Explorer

PranavPurwar:
lmao, why did you make a pr without adding the XML file. Next time before making a pull request, do check that it's complete and test it

Rakshit2:
Please notice I have already added that

PranavPurwar:
@Rakshit2 I have seen that already, I just said that before making a pull request, just check that everything is done and working (atleast the things defined in title)

JavkhlanK:
I'm not sure if this is really needed. There's just too much to change, and I don't think that'd be worth the effort. There also hasn't been too many app crashes recently, right? The bug reporting probably will also include a lot of issues caused by invalid Custom <essential part of Sketchware Pro>, like a `null` value as Custom Block name.

hasrat-ali:
Now we are going to use the Discord webhook as a crash reporter. So I don't think it is less.

But thanks for the contribution.

PR #204, opened 2021-07-29
Feat: Improve ManageJavaActivity

Keller-Zirus:
- Can edit java file from mysc without messing anything.
- Can edit java file not only from package folder
e.g. com/my/any.java
e.g. a.java
- Added ".." to navigate up in the directory tree


JavkhlanK:
I'm not sure what javaM is there for, but it's not needed anymore. The rest of the patches seem great though!

JavkhlanK:
Removing onBackPressed logic wasn't the best tho imo

PR #205, opened 2021-07-29
feat, style: Improve LibraryDownloaderDialog overall

khaled-0:
**feat: LibraryDownloader RepoUrls are now Editable**
The default repo addresses for downloading external libraries logic has been changed so that it can read from ```/.sketchware/libs/repo_map.json```

If the file doesn't exist or is corrupted then it will reset the file to load default repo links.

I know it might be not very useful for now but it can definitely be improved later.

**style: Improved Visual Feedback On LibraryDownloaderDialog**
``` Start , Pause , Resume , Cancel ``` buttons now actually disappear and reappear accordingly when downloading in progress. 

Also EditText now gets disabled when downloading starts and gets re enabled if downloading fails or something.


ilyassesalama:
Are you sure you tested your changes because they seem broken?

khaled-0:
> Are you sure you tested your changes because they seem broken?

I didn't build the app but I did test out the codes in a similar environment, it did work.
But I think I'll review it again

PR #214, opened 2021-08-15
Add a.a.a.IC

Keller-Zirus:
Add a.a.a.IC and
Fix: 
`Can't access Sign/Export when the user is not connected to the internet`

### DIFF
--- a/app/src/main/java/a/a/a/IC.java
+++ b/app/src/main/java/a/a/a/IC.java
@@ -1,0 +1,75 @@
+package a.a.a;
+
+import a.a.a.GC.c;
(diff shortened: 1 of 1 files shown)


hasrat-ali:
@Keller-Zirus did you test your code? Is it working?

Keller-Zirus:
> @Keller-Zirus did you test your code? Is it working?

I tested it with smali.

JavkhlanK:
I'd rather decompile `La/a/a/GC;` and implement changes in its anonymous classes, but until it's available in Java, this is enough.

PR #220, opened 2021-08-22
feat: Add custom More Block return types

Keller-Zirus:
feat: Add custom More Block return types


Keller-Zirus:
WHAT!
`This branch has conflicts that must be resolved`

f6f3eeed524e134d805a1d55d3b85368be697eb6

hasrat-ali:
It doesn't seem necessary to me, because we already have a way for it.
(`moreblockName[type|typeName]` or `moreblockName[type|typeName|typeCode]` )

hasrat-ali:
That would be great.

hasrat-ali:
It will work but without validator, the user can type anything, which will bring new problems.

Then we will have to answer them again and again.

So that's why it would be better for us to use the native method. (`moreblockName[type|typeName`] or `moreblockName[type|typeName|typeCode`] )

So I'm closing this, and thanks for contributing.

PR #223, opened 2021-08-24
feat: add option to show the source code of a single activity.

aliveness-e:
I added a new option to show Java source code of a single activity. No need to open SrcCodeViewer -which can be slow in large projects- just to see a source code of an activity anymore, all you need to do is to open the activity you want to show its source, then click on the "compiler option" button next to the "run" button and select "show source code".

### DIFF
--- a/app/src/main/java/a/a/a/yq.java
+++ b/app/src/main/java/a/a/a/yq.java
@@ -909,6 +909,17 @@ public class yq {
         CommandBlock.x();
         return srcCodeBeans;
     }
+    
+    /**
+     * Get source code of a single Activity
+     */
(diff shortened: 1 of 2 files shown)


hasrat-ali:
@aliveness-e have you tested it?

aliveness-e:
I've only tested the first two commits (the main ones), and they worked fine. but then I felt like some activities (large ones) takes a few seconds to show their sources, so I've adjusted the code to run in a new Thread (the last two commits), but haven't tested those.

hasrat-ali:
Ok, let me try.

JavkhlanK:
It doesn't really feel that much faster than opening SrcCodeViewer, but at least the app doesn't lag while loading sources.

PR #227, opened 2021-08-26
Update design.xml

Ninjacoderhsi:


JavkhlanK:
Since your requested changes don't make much sense and you didn't explain them properly, they'll be closed.

khaled-0:
Ah yes. The real reason why swpro is too slow.
Jbk stop covering it from us. We see right through u

PR #237, opened 2021-09-02
feat: added support to push jar files using git

hosnifraj:
i failed when i try to upload a jar file but looks like i need to change s.th in this file


JavkhlanK:
The .gitignore file isn't for files which can be pushed, it's (sorta) the opposite.

PR #239, opened 2021-09-09
feat, feature: allow using full dependency

PranavPurwar:
Now the user can write full dependency like gradle.
For example: we can now use 
implementation 'something:something:1.0.0'
directly in library downloader dialog


JavkhlanK:
idk the first "feature" doesn't seem too useful to me. Also, it'd leave something like `'com.google.android.material:material:1.4.0'` behind, don't think it'd work with those `'`s.

khaled-0:
why though 🤔

PranavPurwar:
Why: because trimming dependency is pain
yeah using regex might be good but it's slower than replacing

PR #244, opened 2021-09-26
feat: Allow removing orientation tag with inject

aliveness-e:
added the ability to remove "orientation" attribute from LinearLayout.


MikeAndrson:
That's cool and all but what if one wants to remove some other attribute? I think a more generic solution to cover them all would be better.

JavkhlanK:
That really is a hacky way of doing that, especially removing all spaces in the inject attribute, but I guess it's fine until the full `tools:remove="..."` feature is ready.
Also btw, could you PLEASE use commit messages like `feat: Allow removal of orientation tag with inject` as described in the Contribution guidelines next time?

PR #257, opened 2021-10-14
feat: Add Pseudo Logcat Reader

khaled-0:
This PR adds a full fledged ```Logcat Reader``` and Functionality for All(Hopefully) Log Related Stuffs in The Projects Made With.
To enable logcat support in your apps, enable Logcat Option from build settings.
**Make sure you don't keep it enabled in the final builds unless you know what you're doing**

Function Wise (if Logcat Enabled)
Made projects will include a class ```SketchLogger.java```
This Class Uses ```java.lang.Process``` to execute and read each and every (hopefully) logs from the app itself. Then using a BuffereredReader, it Broadcasts the logs one by one with an intent filter ```RECEIVE_NUB_LOGS``` which than gets picked by Sketchware.
Also in order for that ```SketchLogger.java``` class to work one must call ```SketchLogger.startLogging(Context)``` from somewhere in the app. It's usually done automatically in ```SketchApplication.java``` but if for some reason user doesn't use the default application class they must use that method in order for the logcat to work.
**More info can be found in the class itself***

As for Reading The broadcast , Open the ```Logcat Reader``` from Design Menu in Sketchware Pro before opening the application. So you won't miss out the logs of app startup.

Finally. I don't know what else to write anymore.🙏🙏


JavkhlanK (dismissed):

JavkhlanK:
Why's this PR marked as WIP btw?

khaled-0:
i haven't checked the updated DesignDrawer while resolving conflicts. please kindly test that out when you can, should work tho

khaled-0:
These commits are a mess ngl 😔

khaled-0:
cleaned most of messy commits as much as I could.

PR #261, opened 2021-10-18
feat: Adding ElseIf & Else Blocks

khaled-0:
Here's what it looks like 
You can stack as many else if as necessary
Block structure should be like
```
If
else if 
else if
else
```
With as much else if as needed in middle
preview
IDK how's this still not a thing😐 Specially when this is a core function.

### DIFF
--- a/app/src/main/java/mod/hilal/saif/blocks/BlocksHandler.java
+++ b/app/src/main/java/mod/hilal/saif/blocks/BlocksHandler.java
@@ -2620,9 +2620,9 @@ public class BlocksHandler {
             logicEditorActivity.a("c", "whileLoop");
         }
         logicEditorActivity.a("c", "if");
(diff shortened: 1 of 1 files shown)


PranavPurwar:
is this really needed ?

PranavPurwar:
core function does not mean that it's necessary. also custom blocks or asd can be used instead of this

ilyassesalama:
Looks good imo, i always needed something like this in sketchware.

hassanmahfuj:
If you want to make a small app, then the block system is good, but if you work on a big project, the block system s*cks (Lags, Finding, Managing). According to this, "else if" block may also exist.

hassanmahfuj:
Fun Fact : After this conversation I found out that many new blocks have been added by turning on the "Built-in blocks" in the "Mod settings".  I haven't used one yet.

PR #262, opened 2021-10-18
feat: add scrollview to mod settings activity

PranavPurwar:
adds a scrollview in mod settings activity to avoid this 
Screenshot_2021-10-18-22-34-58
> See last option


khaled-0:
Don't mind me. Imma just steal this and fix it real quick.

PR #264, opened 2021-10-26
fix: Replaced Missing Resource Id's With Tags; Fixed Header Type Not Appearing

khaled-0:
Nothing To Put Here. Changes have been tested and It Works


khaled-0:
Can you merge this already please.
Otherwise i need to make those changes every time i fetch from upstream

PR #266, opened 2021-10-30
feat: remove useless shared collection from logic editor's moreblock section

PranavPurwar:


khaled-0:
Remove others too. From design activity and anywhere else👍

PR #268, opened 2021-11-06
feat: Backup Projects In Their Respected Folders With Version Name & Code Included

khaled-0:
Title Is Self Explanatory. 
But in Case Anyone's Wondering . 
Project backups will now be grouped inside the named folder of project
and filename will include the version name & code (including newly added version text)
Format: ``ProjectName v1.0 (com.whatever.projectname, 110) 2021-12-31T125827.swb``
Here's the Expected Behaviour
image


PranavPurwar:
Ok, but that file name looks too messed, maybe remove package name and replace full date and time to something like “12:35 8-11”

khaled-0:
> Ok, but that file name looks too messed, maybe remove package name and replace full date and time to something like “12:35 8-11”

There was a vote 🐸

PranavPurwar:
ok nub poll, just saw the chat in discord

PR #272, opened 2021-11-20
fix:Make DebugActivity text actually selectable,Cleanup unnecessary null checks

khaled-0:
findViewById after dialog.show when?
also why null checks?

### DIFF
--- a/app/src/main/java/com/besome/sketch/DebugActivity.java
+++ b/app/src/main/java/com/besome/sketch/DebugActivity.java
@@ -21,13 +21,8 @@ public class DebugActivity extends Activity {
                         .setMessage(error)
                         .setPositiveButton("End Application", (dialog1, which) -> finish())
                         .create();
-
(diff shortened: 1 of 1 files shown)


JavkhlanK:
turns out that selectable stack trace's still not working, haha

khaled-0:
> turns out that selectable stack trace's still not working, haha

how fast was that, i wal also updating asset/debug.java
anyway

JavkhlanK:
I don't get why you would submit PRs without fully finishing them first

khaled-0:
((TextView) dialog.findViewById(android.R.id.message)).setTextIsSelectable(true);
image

khaled-0:
> I don't get why you would submit PRs without fully finishing them first

i didint think u would merge it instantly, 
please update the other file urself

JavkhlanK:
That file looks poorly formatted, so I'll definitely clean it up & also fix that as well

PR #277, opened 2021-12-04
feat: You can now modify the filename format of project backups

khaled-0:
How cool is that, I know right.
PR Is Self Explanatory, Or look at the images
**Input:**
image
**Output:**
image


DerGoogler:
Add simple highlight, to make understanding easier

PR #273, opened 2021-11-22
feat: improve build speed

PranavPurwar:
By stopping the unnecessary merging of dex files during debug build

PR #278, opened 2021-12-06
better version of #273

PranavPurwar:


PranavPurwar:
DO NOT SAY THAT COPYING LIBS TO EXTERNAL STORAGE IS UNNECESSARY, BECAUSE IF I DON'T THEN THE APP WILL CONTAIN THEM AS "MULTIDEX-XXX.DEX" WHICH WILL CRASH THE APK

tyron12233:
whats with the all caps

PranavPurwar:
NOTHING

PranavPurwar:
merge this and release beta 7 when

PranavPurwar:
another tip: use Incremental compilation from {@package mod.tyron.compiler} before releasing next beta

JavkhlanK:
Behavior like "screaming" with caps lock on is not welcome here. The same applies to the attitude you have, especially when phrased like "Don't say that […] because if I don't […]". Since you already got warned before, measures will be taken in an appropriate manner.

Please stop being this toxic, or else, you will be banned permanently from collaborating in this repository.

PR #286, opened 2021-12-16
add android ci

PranavPurwar:


khaled-0:
Description 😔
Title 😔

khaled-0:
feat,misc: Add GitHub workflow integration to auto generate latest apk

Idk description 😔

ilyassesalama:
I bet the compilation will fail. Android CI in Sketchware Pro repository is useless.

JavkhlanK:
Why would compilation fail? Looks like it would do the same as building a debug APK with Android Studio.

And about it being useless, contribution would be possible without needing to own a desktop with Gradle + the Android SDK on it, which would enable (easy) pure mobile-only development again.

ilyassesalama:
Fair enough.

PranavPurwar:
> I bet the compilation will fail. Android CI in Sketchware Pro repository is useless.

gimme $1 if you lose the bet 😳

khaled-0:
Ha

PranavPurwar:
shut up

PR #287, opened 2021-12-17
Firebase without Appcompat

Did-Dog:
A main problem in sketchware ,can't use firebase without approbation.
I removed the connection of appcompat and firebase.Now we can use firebase without approbation.

### DIFF
--- a/app/src/main/java/com/besome/sketch/editor/manage/library/ManageLibraryActivity.java
+++ b/app/src/main/java/com/besome/sketch/editor/manage/library/ManageLibraryActivity.java
@@ -172,9 +172,9 @@ public class ManageLibraryActivity extends BaseSessionAppCompatActivity implemen
                     initializeLibrary(libraryBean);
                     if (libraryBean.useYn.equals("Y") && !compatLibraryBean.useYn.equals("Y")) {
                         libraryBean = compatLibraryBean;
-                        libraryBean.useYn = "Y";
+                        libraryBean.useYn = "N";
                         initializeLibrary(libraryBean);
-                        showFirebaseNeedComaptDialog();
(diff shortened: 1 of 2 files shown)


PranavPurwar:
i think it already works without that, I've used firebase without appcompat in many projects (earlier)

ilyassesalama:
This is interesting, needs a detailed test though.

Did-Dog:
> This is interesting, needs a detailed test though.

Did-Dog:
> > This is interesting, needs a detailed test though.
How can I compile Sketchware Pro's source into an APK file

PranavPurwar:
clonr the repository on android studio, click Sync Gradle and click run/debug

Did-Dog:
> clonr the repository on android studio, click Sync Gradle and click run/debug

Ok

PranavPurwar:
or if you're in mobile ->
add this code to android config in bulid.gradle and test 

jackOptions {
      enabled true
    }
  }
  compileOptions {
    sourceCompatibility JavaVersion.VERSION_1_8
    targetCompatibility JavaVersion.VERSION_1_8
  }

JavkhlanK:
That looks like an interesting change, I'll test that out soon as well. By the way, you should rather remove code entirely instead of commenting it out, as the entire history (including the old unchanged version) is available thanks to Git.

JavkhlanK:
Unfortunately, Sketchware didn't lie that Firebase depends on AppCompat. Looking at the dependencies of just `com.google.firebase:firebase-common:19.0.0` (exactly the version built-in Sketchware Pro), you can find `com.google.android.gms:play-services-basement:17.0.0` which depends on various AppCompat (AndroidX, actually) libraries, so not including them in the APK is **not safe**.

PR #291, opened 2021-12-25
fix: Check for storage perms before reading config

khaled-0:
If no storage perms , don't try to read for custom config,
Fixes startup crash

### DIFF
--- a/app/src/main/java/com/besome/sketch/MainActivity.java
+++ b/app/src/main/java/com/besome/sketch/MainActivity.java
@@ -3,6 +3,7 @@ package com.besome.sketch;
 import android.Manifest;
 import android.app.AlertDialog;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.graphics.Color;
 import android.net.Uri;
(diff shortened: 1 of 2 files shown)


JavkhlanK:
I think changing ConfigActivity's logic to not try to read Mod Settings if no storage permissions have been granted and/or handle `null` HashMaps correctly would be a better solution.

khaled-0:
> I think changing ConfigActivity's logic to not try to read Mod Settings if no storage permissions have been granted and/or handle `null` HashMaps correctly would be a better solution.

cant check for perms without passing context, which seems unnecessary if ,
lets try to handle null then

khaled-0:
Ha, much better now

JavkhlanK:
> > I think changing ConfigActivity's logic to not try to read Mod Settings if no storage permissions have been granted and/or handle `null` HashMaps correctly would be a better solution.
> 
> cant check for perms without passing context, which seems unnecessary if ,
> lets try to handle null then

It'd be possible to use `SketchApplication#getContext()` for a Context.

khaled-0:
> > > I think changing ConfigActivity's logic to not try to read Mod Settings if no storage permissions have been granted and/or handle `null` HashMaps correctly would be a better solution.
> > 
> > 
> > cant check for perms without passing context, which seems unnecessary if ,
> > lets try to handle null then
> 
> It'd be possible to use `SketchApplication#getContext()` for a Context.

Not necessary anymore , because null check serves the purpose.

PR #296, opened 2021-12-29
fix crash on low api

PranavPurwar:
Context#getColor method was added in api 23


JavkhlanK:
Using `ContextCompat#getColor(Context, int)` might be better for that, as it checks the API version and calls the newer method if supported.

JavkhlanK:
This got resolved with 7d2b48eb0784519947baa30f9530c4ef4abb95d1 already.

PranavPurwar:
ik

PR #325, opened 2022-02-06
feat: Create Vastly Improved SrcViewerActivity.java

khaled-0:
-Asynchronous source generation
-Better code viewer (Sora editor)
-Preserves the original looks

Feels 50% faster ngl 
image


JavkhlanK:
That lowercase class name though

Issue #323, opened 2022-02-05
Error showing source code for custom views


brunodavi:
When I go to `custom_view` > `settings_button` > `Show source code` it shows an error when viewing the `.xml`

> Not sure if it could be because I restored the project from a backup (.swb)

<details>
<summary>Screenshots</summary>

![sketchware_select_custom_view_aaa.jpg](https://user-images.githubusercontent.com/48166556/152654019-09d0f7cb-2fd0-42ff-a851-ce28da3d9906.jpg)

![sketchware_click_show_source_code.jpg](https://user-images.githubusercontent.com/48166556/152654143-88cbfbae-fae4-48a7-be80-06481d934096.jpg)

![sketchware_error_show_source_code.jpg](https://user-images.githubusercontent.com/48166556/152654179-a399452b-ca32-4be9-9613-6bc35cf85dec.jpg)

> Here's a sequence of what happens

</details>

JavkhlanK:
Thanks for the report! Was able to reproduce, might try to fix before v6.4.0-rc02's release.

JavkhlanK:
Fixed by #327.

JavkhlanK:
Wait, I'll keep this issue open until v6.4.0-rc02 gets released.

JavkhlanK:
Fixed with [v6.4.0-rc02](https://github.com/Sketchware-Pro/Sketchware-Pro/releases/tag/v6.4.0-rc02).

PR #327, opened 2022-02-07
fix: Quick source viewer for custom xml works now.

khaled-0:
Also the command blocks now get applied in xml quick source
-Sideeffect: Relatively long loading time for xml sources
image


khaled-0:
fixes issue #323

PR #334, opened 2022-02-23
Update build.gradle

ArjunaKumarMohanta:


JavkhlanK:
Please commit that to your own fork. Ignoring lint errors is not needed when we build Sketchware Pro's sources.

PR #337, opened 2022-03-03
feat:refactor: AppComponents Sora editor, Quick manifest source

khaled-0:
Use sora editor if Legacy editor not enabled in settings
Delete old AppComponentsDialog class

you may or may not want the quick(so called) source
It's almost same speed as the full source generation.

And some cleanups in ManifestManager.


JavkhlanK:
It would be better if the old AppComponents dialog would still get shown if the Use legacy Code Editor Mod setting is on.

khaled-0:
> It would be better if the old AppComponents dialog would still get shown if the Use legacy Code Editor Mod setting is on.

It does doe, uses legacy editor but not in dialog
And the legacy editor needs a cancel & Save button anyway, so it should be fine soon

JavkhlanK:
Oh I missed that, thanks.

PR #342, opened 2022-03-08
Remove SketchwareUtil & SketchApplication classes

kimoandroid:
hello, can you please remove these two files from sketchware SketchwareUtil & SketchApplication
because i tried too much to remove them but failed


JavkhlanK:
What does this Pull request change? Please open an Issue instead.

PR #384, opened 2022-05-08
Update ExtraPaletteBlock.java

Arashvscode:
Announce the Dialog Progress Background


JavkhlanK:
Firstly, please open only one Pull request for each feature. #383 and #384 are together a single feature.

Secondly, there will be no more built-in Blocks in Sketchware Pro. Users should use Custom Blocks instead.

And thirdly, it doesn't look like you ever tested your changes out. There are two typos that make the Block not even show up correctly.

Arashvscode:
> 

Ok my bro and sorry 😅😅

PR #397, opened 2022-05-15
misc: Changed the splash screen slightly

khaled-0:
gradient color now uses the one from logo
increased logo size instead of getting from mipmap
removed hardcoded delay.

and ZipSigner because that warning was very much bothering


khaled-0:
Certified by image btw

JavkhlanK:
I've never seen `sp` used for dimensions other than text size before lol

khaled-0:
> I've never seen `sp` used for dimensions other than text size before lol

dp has no difference at all but changed anyway

JavkhlanK:
Now it looks weirdly off in my opinion. I think we should remove the useless activity entirely, and instead, use a splash screen with resource themes.

JavkhlanK:
Nevermind, was due to Android 12 not liking outdated libraries.

hasrat-ali:
> Now it looks weirdly off in my opinion. I think we should remove the useless activity entirely, and instead, use a splash screen with resource themes.

After this [commit](https://github.com/Sketchware-Pro/Sketchware-Pro/commit/803de939f7f6e24c5aa2b884502912ecf27ccf34) now there is no need for it (#397).

PR #402, opened 2022-05-16
feat: add "Auto complete symbol pair" option to CodeEditor.

Raival-e:


AlQutby:
- Add a writing box, to write the application version freely

2-Add A list of distinctive, to appear where the projects that worked for her star franchise, which work on them continuously, and access them quickly at any time you want to complete.


3-  Direct Code Editor

4- Turn the app from apk to AAB

PR #418, opened 2022-05-25
Update zipalign-java

Iyxan23:
This pull request updates the zipalign-java jar to the current latest commit of `Iyxan23/zipalign-java` at `96fd6b30c928015847264b27b0d2dd407f3e8895`. It also updates Sketchware Pro to use the faster method.

Aligning zip is now considerably faster than it previously was.

Using the previous sketchware pro version, aligning a ~187KiB apk takes about `3494ms`. This new update cuts down that time to just only `684ms`.


17606657524:
Flash back

17606657524:
On the mobile phone, I use the flash back version.

Iyxan23:
The build is crashing due to the latest commit being buggy, please use this build instead that's based off of the latest rc with the same update commits applied: https://github.com/Iyxan23/Sketchware-Pro/actions/runs/2384429111

Iyxan23:
sigh, could've used the `zipalign-update` branch as the incoming branch instead

anthony-china:
ok

anthony-china:
We look forward to optimizing the packaging speed

PR #436, opened 2022-06-03
Resolve unexpected comment.

LetsDoNothing:


JavkhlanK:
That's a TODO comment, something tracking features to implement. Would be better if the comment got replaced by the feature it mentions, not nothing.

hasrat-ali:
Who should explain to this brother that this is not an Unexpected comment.

hasrat-ali:
Sorry, but this should be closed.

PR #441, opened 2022-06-06
refactor,fix: Change how application crashes are handled

khaled-0:
Improve DebugActivity and UncaughtException Behavior.
AlarmManager fails on most of devices, So let's not use it.

Saving the crash log to a SharedPreference and showing it on next startup is most likely a better approach.

### DIFF
--- a/app/src/main/assets/debug/DebugActivity.java
+++ b/app/src/main/assets/debug/DebugActivity.java
@@ -1,76 +1,63 @@
 package <?package_name?>;
 
+import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.AlertDialog;
+import android.content.Context;
 import android.content.DialogInterface;
-import android.content.Intent;
 import android.os.Bundle;
-import android.util.Log;
 import android.widget.TextView;
 
 public class DebugActivity extends Activity {
-
-    private String[] exceptionTypes = {
-            "StringIndexOutOfBoundsException",
-            "IndexOutOfBoundsException",
-            "ArithmeticException",
-            "NumberFormatException",
-            "ActivityNotFoundException"
-    };
-
-    private String[] exceptionMessages = {
-            "Invalid string operation\n",
-            "Invalid list operation\n",
-            "Invalid arithmetical operation\n",
-            "Invalid toNumber block operation\n",
-            "Invalid intent operation"
-    };
+    private AlertDialog dialog;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
-        Intent intent = getIntent();
-        String errorMessage = "";
-        String madeErrorMessage = "";
-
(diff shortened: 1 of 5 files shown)


app/src/main/assets/debug/DebugActivity.java
```diff
@@ -1,76 +1,63 @@
...
+        context.getSharedPreferences("SK_CRASH_LOG", Context.MODE_PRIVATE).edit().putString("errorLog", errorLog).commit();
+    }
+
+    public static void clearErrorLog(Context context) {
+        context.getSharedPreferences("SK_CRASH_LOG", Context.MODE_PRIVATE).edit().remove("errorLog").apply();
+    }
+
+    public static String getLastSavedErrorLog(Context context) {
+        return context.getSharedPreferences("SK_CRASH_LOG", Context.MODE_PRIVATE).getString("errorLog", "");
+    }
```
JavkhlanK, line 62: I'm not sure if these should be added. Also, using the application context stored by the default application class might be better.

app/src/main/assets/debug/DebugActivity.java
```diff
@@ -1,76 +1,63 @@
...
-            "ActivityNotFoundException"
-    };
-
-    private String[] exceptionMessages = {
-            "Invalid string operation\n",
-            "Invalid list operation\n",
-            "Invalid arithmetical operation\n",
-            "Invalid toNumber block operation\n",
-            "Invalid intent operation"
-    };
```
JavkhlanK, line 27: I really don't think these simplified error messages should be removed.

JavkhlanK (changes requested):

app/src/main/assets/debug/DebugActivity.java
khaled-0: It's generally a better idea and practice to pass the context along. And about methods,
Bro sketchware is using a whole activity for only showing a single TextView, so adding some helper methods instead of putting them in application class doesn't make sense.

app/src/main/assets/debug/DebugActivity.java
khaled-0: It doesn't help anymore because of the bizzare custom blocks people uses these days.
They were helpful in vanila sw sure, but I don't think the extra stuffs are needed anymore.
But you can add them back if you think those are necessary.
Imo they're useless.

hasrat-ali: @khaled-0 what would you say now?

khaled-0: you're intentionally crashing the app 
it's causing to crash on every startup.

I dont think this behaivour can be avoided with what this pr does for now.
unless there's a way to not continue the app if there's a crash

khaled-0:
I'll close this as it's not generally better than current either as is seems.

PR #442, opened 2022-06-06
Error' bug in font assest

taikucing1:
Rca 02_03 error' not visible font ,where ttf to pmv


hasrat-ali:
> Rca 02_03 error' not visible font ,where ttf to pmv

If you have any problem pls open issue instead.

PR #487, opened 2022-07-01
Update SrcCodeEditor.java

Arashvscode:
Prove a counting liner and present a more beautiful effect!


JavkhlanK (changes requested):
I feel like this should instead be an optional code editor setting.

nonzzzz:
`setPinLineNumber` enabled by default.

PR #494, opened 2022-07-08
Update KotlinCompilerBridge.java

aladinim:
I am checking this in every device. So the Kotlin was working well on v27 but now it stopped. Because of HeyStudio's idea i guess.


aladinim (commented):
,

JavkhlanK:
I'm not sure what this is supposed to do, but thanks for making me notice how that useless method & its checks are.

MikeAndrson:
Got referred to as "Hey Studios" after so much time 🥲 Really takes you back

MikeAndrson:
Also yeah those are redundant since kt feature is on a different flavor anyway

PR #516, opened 2022-07-26
Update proguard to support java 11

PranavPurwar:
The currently used version doesn't support Java 11 which is now availablw for use


JavkhlanK:
Why did you add both base and core?

NiceSapien:
Noice

PranavPurwar:
.

Arashvscode:
> .

حتی فکرشم نکن

PHANTOM-XOTIC:
دیوانه

PR #532, opened 2022-08-12
New feature

Arashvscode:
The Google Material library has been updated to 1.6.1, the Gson library has also been updated to 2.9.1, it is better to keep the line pin fixed by default in the Java editor code, it shows a more beautiful face.  The version of such libraries of the program itself is also updated


app/src/main/java/mod/hey/studios/code/SrcCodeEditor.java
```diff
@@ -360,6 +361,7 @@ private void initializeLogic() {
         setTitle(getIntent().getStringExtra("title"));
 
         editor.setTypefaceText(Typeface.MONOSPACE);
+        editor.setPinLineNumber(true);
```
JavkhlanK, line 364: Like I said in #487, [this'd be better as Code Editor setting](https://github.com/Sketchware-Pro/Sketchware-Pro/pull/487#pullrequestreview-1033561995).

app/build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 32
     buildToolsVersion "30.0.3"
```
JavkhlanK, line 8: You changed the compile SDK version, but not the build tools version. This is not good practice and gives off warnings in Android Studio.

JavkhlanK (changes requested):
Some of these seem okay, but you added so many changes into a single Pull request, which forces us to accept all, or none. Since some are unwanted at this stage, I can't merge this in.

I'm also gonna use this to add a comment on updating of libs.zip and dexs.zip, since GitHub doesn't allow commenting on binary changes. Sorry, can't accept that. It's a bit like accepting an untrusted random executable, I don't trust that.

app/build.gradle
Arashvscode: OK, I will restore it to the previous state

Arashvscode:
> 

I added the official libraries, the same thing that was in Google, it doesn't need trust. This shows that you don't even trust Google. Meanwhile, I tested the program, there is no problem, and even my project compiles and runs well.  thanks

JavkhlanK:
Google doesn't provide any official DEX files of their libraries, nor them packaged in libs.zip as in Sketchware Pro. You are not Google, that's what I meant.

Arashvscode:
> Yes, actually they are all jar, I converted them to DEX with D8 tools and added them to your service, it's easy, I guarantee you that there are no errors. 😁😁😁😅

Arashvscode:
https://user-images.githubusercontent.com/101287880/184471829-82c27053-f9ca-4173-b948-ea8f7d0e0742.mp4



Very good, this is the result of the update of Gson and Google Material. I tested the sw pro program from my own fork. You can see that it worked. Gson and Google Material are much better than before, thank you.

PR #537, opened 2022-08-14
Added executable shell script upon startup

onyxfluxx:
Anyone who is good with shell scripting could put this to good use. Usage could include automatic restore of projects, local libs, custom blocks, components, and more. works with root and without.  Can be turned on or off in mod settings, and shell script must be manually made inside root of internal storage. sketchwarescript.sh would be the name of the script.


onyxfluxx:
making process async so it doesnt overload main thread.

PR #580, opened 2022-09-14
Create gradle-wrapper-validation.yml

ghost:


JavkhlanK:
Great change, though undocumented what it does. Next time, please add a commit message explaining the change, as well as Pull request title and description.

What also would be great is for that Actions workflow to only run on changes to the Gradle wrapper, not on all Pull requests.

JavkhlanK:
What?

PR #584, opened 2022-09-19
update firebase version to 19.3.0

AdesonBezerra:
With firebase version 19.3.0 we will be able to use increment values ​​to send data to the database easily and you can also use the write timestamp

This change takes no more than 3 lines of code and won't mess up old projects before this change.

In this way we will be able to use the methods:
ServerValue.TIMESTAMP
ServerValue.increment(0)

Example:
send.put("balance", ServerValue.increment(1));
send.put("timestamps", ServerValue.TIMESTAMP);
Db.path("users/uid).update(send);

import com.google.firebase.database.ServerValue; 

Note: I already made the modification using the latest version and it worked without problems, follow the video of how I did it to edit the mod 
https://drive.google.com/file/d/1PdvlJ_PlPtWpQwMGXjFlEXYXO5Iwp3N6/view

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -9,29 +9,20 @@ android {
 
     def gitCommitHash = "git rev-parse HEAD".execute().text.trim()
 
-    def getCommitHash = { ->
-        def stdout = new ByteArrayOutputStream()
-        exec {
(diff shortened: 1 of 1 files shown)


JavkhlanK:
Not a pull request.

PR #606, opened 2022-10-09
feat: added developer tools menu in designer drawer

BhavyUkani:


JavkhlanK:
Why there though? You could just open an Event and go to Developer Tools from there.

PranavPurwar:
close

PR #744, opened 2023-02-04
chore: Remove unwanted files

TheRemakerMan:
I think that these files were automatically generated by the decompiler.

I also think that this pull request will make life easier for contributors. There will be no extra unused files.


hasrat-ali:
👍

hasrat-ali:
Btw, how you mapped those layout/XML files are duplicate or unused?

TheRemakerMan:
> Btw, how you mapped those layout/XML files are duplicate or unused?

Manually searched for each file. If this file is not used in files along the path: `app/src/main/java/...` then delete the file.

TheRemakerMan:
I think there will be errors if I delete these files (some libs in `app/libs` also use above removed files) so I'm closing this pull request.

JavkhlanK:
A more fine-grained cleanup of resources added by AndroidX AppCompat and/or MDC would be better, for example those abc_… drawables.

PR #776, opened 2023-02-23
Lib Version Update as same as Sketchware pro

DevVigilante:
com.google.android.material > 1.6.1
androidx.appcompat > 1.4.0

Exported source code using older lib than Sketchware pro run fix.


hasrat-ali:
LGTM

hasrat-ali:
Looks good to me

PR #777, opened 2023-02-23
Update Lx.java

moderGamer:
feat: add last Version Gson and Androidx(AppCompat)

### DIFF
--- a/app/src/main/java/a/a/a/Lx.java
+++ b/app/src/main/java/a/a/a/Lx.java
@@ -59,7 +59,7 @@ public class Lx {
                 "implementation fileTree(dir: 'libs', include: ['*.jar'])\r\n";
 
         if (metadata.g) {
-            content += "implementation 'androidx.appcompat:appcompat:1.4.0'\r\n" +
+            content += "implementation 'androidx.appcompat:appcompat:1.6.0'\r\n" +
                     "implementation 'com.google.android.material:material:1.6.1'\r\n";
         }
 
@@ -88,7 +88,7 @@ public class Lx {
         }
 
         if (metadata.isGsonUsed) {
-            content += "implementation 'com.google.code.gson:gson:2.8.7'\r\n";
+            content += "implementation 'com.google.code.gson:gson:2.9.0'\r\n";
         }
 
         if (metadata.isHttp3Used) {


moderGamer:
> 

what?

PranavPurwar:
Sketchware doesn't use gradle, this is useless

moderGamer:
> 

I know that I did Android Studio to output the project, who knows, maybe Gradle will be supported in the future

TechnicalStudioDeveloper:
> > 
> 
> I know that I did Android Studio to output the project, who knows, maybe Gradle will be supported in the future

I know because my project gives error because of wrong gradle version in Android Studio that why I updated it.

TechnicalStudioDeveloper:
bro! The version you updated is more higher than the Sketchware compile in run build.I think that is why hasrat-ali marked this as invalid.

moderGamer:
> 

Maybe, but Hasrat should update the Androidx library, can't individual users use a custom library?!
@hasrat-ali  @TechnicalStudioDeveloper

hasrat-ali:
Pls don't mention anyone.
Will see when we update it.

TechnicalStudioDeveloper:
> > 
> 
> Maybe, but Hasrat should update the Androidx library, can't individual users use a custom library?! @hasrat-ali @TechnicalStudioDeveloper

Changing a single line of code to change lib version is not enough in my opinion because updating a lib can break old projects and to prevent them from throwing error Sketchware pro should make it compatible with new that version which is hard in my opinion.

TechnicalStudioDeveloper:
> > 
> 
> Maybe, but Hasrat should update the Androidx library, can't individual users use a custom library?! hasrat-ali @TechnicalStudioDeveloper

yeah! from user view it is necessary that is why there is **Exclude built in lib**...

moderGamer:
> 

Great, so there is nothing left to say, I can also update the libraries, but I'm afraid it will be rejected again

TechnicalStudioDeveloper:
> > 
> 
> Great, so there is nothing left to say, I can also update the libraries, but I'm afraid it will be rejected again

😂😹 That's why I am also afraid and made a much small contribution.

moderGamer:
> 
Yes, it is better for Hasrat Ali or jbk0 to update it because many of the old methods of Android X are obsolete, for example, locking ViewPager, which was removed or changed in version 1.6.1.

JavkhlanK:
> > 
> 
> Great, so there is nothing left to say, I can also update the libraries, but I'm afraid it will be rejected again

That's true because I can't trust everyone to do such a complicated task properly.

moderGamer:
> 

Yes, you are absolutely right, there is no doubt about it, so I leave the update of important and main libraries to you, just try to add it to the latest available version.

and add new Block 

```java
recyclerview1.getAdapter().notifyDataSetChanged();
```

I think you forgot this code, make it public to refresh the data

moderGamer:
> 

Done.

PR #779, opened 2023-02-24
feat: Replace "Select MoreBlock type" with "Return".

Alejandromarmilich:
This will be more real and let's take advantage of the return blocks in Control, I think it makes more sense.

### DIFF
--- a/app/src/main/res/layout/make_block_layout.xml
+++ b/app/src/main/res/layout/make_block_layout.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:background="#ffffffff"
@@ -95,7 +96,7 @@
                     android:layout_marginLeft="16dp"
                     android:layout_marginTop="4dp"
                     android:layout_marginRight="8dp"
-                    android:text="Select Moreblock type"
+                    android:text="Return: "
                     android:textColor="@color/color_primary"
                     android:textSize="14sp"
                     android:textStyle="bold" />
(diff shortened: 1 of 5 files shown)


app/src/main/res/values-land/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="add">ADD</string>
+    <string name="name">name</string>
+    <string name="parameter_without">parameter (without %)</string>
+</resources>
```
JavkhlanK, line 6: Why a "different" value for landscape mode? If you don't specify it again, the default value will be used anyway.

app/src/main/res/values-w820dp-v13/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="add">ADD</string>
+    <string name="name">name</string>
+    <string name="parameter_without">parameter (without %)</string>
+</resources>
```
JavkhlanK, line 6: Same as in values-land/strings.xml here.

app/src/main/res/values-xhdpi-v4/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="add">ADD</string>
+    <string name="name">name</string>
+    <string name="parameter_without">parameter (without %)</string>
+</resources>
```
JavkhlanK, line 6: Also same as in values-land/strings.xml.

app/src/main/res/layout/make_block_layout.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
```
JavkhlanK, line 3: This namespace declaration is unused.

JavkhlanK (changes requested):
There's a few issues, but in general, I don't see the need to change "Select Moreblock type" to "Return: ".

PR #803, opened 2023-03-10
feat : Project compiling notification.

DevKumar00001:


TechnicalStudioDeveloper:
XRecorder_10032023_131819
XRecorder_10032023_131947

khaled-0:
Well it's not actually building in the background
the activity/build may get killed by system when you're on background
because the build process is actually not in background thread/service

So this doesn't matter

TechnicalStudioDeveloper:
> Well it's not actually building in the background the activity/build may get killed by system when you're on background because the build process is actually not in background thread/service
> 
> So this doesn't matter

I mean to say that when Sketchware is not open in foreground it is paused for building app so notification will only appear when app is not foreground.

TechnicalStudioDeveloper:
fixed error : [error](https://discord.com/channels/790686719753846785/1042132489084338236/1084527722560233502)

TechnicalStudioDeveloper:
Trying to fix : [Error](https://discord.com/channels/790686719753846785/1042132489084338236/1084856974149169194)

DereckySany:
I liked the addition. but I see no reason to use androidx.core.app.NotificationCompat.Builder
I prefer to use the native libraries as much as possible, so as not to generate incompatibility or the use of one more library, and you can use Notification.Builder natively.

TechnicalStudioDeveloper:
> I liked the addition. but I see no reason to use androidx.core.app.NotificationCompat.Builder I prefer to use the native libraries as much as possible, so as not to generate incompatibility or the use of one more library, and you can use Notification.Builder natively.

IMG_20230412_103023
But it working on Api-23 and I am not sure but it will work on lower devices also.

TechnicalStudioDeveloper:
It only have one bug that I can't fix is If the project is running and app get destroyed from recent screen it will keep showing a notification.

JavkhlanK (changes requested):
I don't think sending notifications about build status while the actual build process itself isn't guaranteed to not get killed (or at least given higher importance than other apps) while it's not in the foreground is worth implementing. It could actually make users believe the app can compile in the background, and then think compiling stopping while multi-tasking being a bug.

TechnicalStudioDeveloper:
> I don't think sending notifications about build status while the actual build process itself isn't guaranteed to not get killed (or at least given higher importance than other apps) while it's not in the foreground is worth implementing. It could actually make users believe the app can compile in the background, and then think compiling stopping while multi-tasking being a bug.

Then it was better earlier that it only shows notification when build is failed or ready to install.

TechnicalStudioDeveloper:
> > I don't think sending notifications about build status while the actual build process itself isn't guaranteed to not get killed (or at least given higher importance than other apps) while it's not in the foreground is worth implementing. It could actually make users believe the app can compile in the background, and then think compiling stopping while multi-tasking being a bug.
> 
> Then it was better earlier that it only shows notification when build is failed or ready to install.

@JavkhlanK

TechnicalStudioDeveloper:
@PranavPurwar If it just show a notification when build is ready then there will no problem.
A install button appear when build ready but if the app is killed then notification will not appear.And same for the failure(If compile issue occured).

TechnicalStudioDeveloper:
> @PranavPurwar If it just show a notification when build is ready then there will no problem. A install button appear when build ready but if the app is killed then notification will not appear.And same for the failure(If compile issue occured).

[This commit](https://github.com/Sketchware-Pro/Sketchware-Pro/pull/803/commits/080efa9facd977cf9a28c083169edb8cb67b87bc) work perfectly as I said @PranavPurwar

PranavPurwar:
and what if some compile issue occurs

TechnicalStudioDeveloper:
> and what if some compile issue occurs

It says app build failed in a notification and a button appear "Show compile log"
and it redirect to error activity.
Till that commit all work as I said

TechnicalStudioDeveloper:
> > and what if some compile issue occurs
> 
> It says app build failed in a notification and a button appear "Show compile log" and it redirect to error activity. Till that commit all work as I said

@PranavPurwar wdyt.
It was not better earlier?

Issue #401, opened 2022-05-16
Youtube Player


whazza1983:
Hello the youtube player is not working.
My app keeps crashing

JavkhlanK:
We're aware of that bug, but aren't really sure why that's happening.

To get to the details, the APK is crashing due to the application class not getting found somehow. Opening the APK's DEX files with e.g. MT Manager shows that the classes are there though. That's as far as I've analyzed that problem so far.

AlQutby:
كيف استطيع التواصل مع مطورين سيكتشوير برو؟

في الاثنين، 16 ماي 2022 10:53 م Javkhlan K. ***@***.***> كتب:

JavkhlanK:
@AlQutby, if you want to contact the "developers" of Sketchware Pro, use this GitHub repository, or join the official Discord server and ask in, for example, the channel #general.

AlQutby:
لدي اقتراحات مهمة داخل سيكتشوير وضرورية لاي مطور اريد اقتراحها لهمSent from my Verizon, Samsung Galaxy smartphone
-------- الرسالة الأصلية --------من: "Javkhlan K." ***@***.***> التاريخ: 16‏/5‏/2022  11:03 م  (GMT+03:00) إلى: Sketchware-Pro/Sketchware-Pro ***@***.***> نسخة: AlQutby ***@***.***>, Mention ***@***.***> الموضوع: Re: [Sketchware-Pro/Sketchware-Pro] Youtube Player (Issue #401) 
@AlQutby, if you want to contact the "developers" of Sketchware Pro, use this GitHub repository, or join the official Discord server and ask in, for example, the channel #general.

—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/Sketchware-Pro/Sketchware-Pro/issues/401#issuecomment-1128081529",
"url": "https://github.com/Sketchware-Pro/Sketchware-Pro/issues/401#issuecomment-1128081529",
"name": "View Issue"
},
"description": "View this Issue on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

User0790:
YoutubePlayerView works great in sketchware v6.4.0 beta 6
[...]

Issue #807, opened 2023-03-13
YouTube player not working


jackw2022:
When the app has lot of classes.dex it crashes on opening but when i merge classes.dex it works..
please add option to enable or disable multidex .
thank you for your efforts

JavkhlanK:
Duplicate of #401.

Issue #780, opened 2023-02-24
Bug In Sketchware v6.4.0 test build 5


HasnatFaiz:
Youtube Player Widget is not working without signing app.

JavkhlanK:
Duplicate of #401.

TechnicalStudioDeveloper:
Fixed in [pull request](https://github.com/Sketchware-Pro/Sketchware-Pro/pull/810)

PR #810, opened 2023-03-16
Fix and Feat

elfilibusterismo:
Fix : YouTubePlayerView
by updating kotlin stdlib to 1.6.0


JavkhlanK:
1. Please keep different changes/fixes in different Pull requests. Closing #803 just to include its changes in here was a bad decision. What if #803's original changes are ready to be merged, but the new ones you added here not? All changes would have to wait before they can be merged.
2. Also don't comment *the same message in three different Issues, some of them even closed*, that's spam. If you really want to reference them in your Pull request, add something like this to its description: "Fixes #401, #780, #807". Better to only mention the open Issue though, the others are duplicates of it.
3. When modifying built-in libraries, at least also tell *what you did to fix the issue*. We need to also be able to reproduce the .dex file(s) you compiled, otherwise we can't be sure if it contains malicious code or not.
4. I hope this is redundant to mention, but I'll have a look at this PR when I've got the time.

elfilibusterismo:
Oh, yah. Forgot to tell. I updated the kotlin stdlib to 1.6.0 and stdlib-jdk7 to 1.6.0 I still used the old name "kotlin-stdlib-1.4.30-M1-release-152" same with jdk7. I still used the the same name because if i changed the name to 1.6.0 I'm worried to it may cause a conflict on other feature because it uses the kotlin-stdlib-1.4.30-M1-release-152 name. Like exclude Library feature.

elfilibusterismo:
Sorry this is my first time. I just ask TechnicalStudioDeveloper to help me commit and make a pull request.

JavkhlanK:
`org.jetbrains.kotlin:kotlin-stdlib:1.6.0` seems to depend on `org.jetbrains.kotlin:kotlin-stdlib-common:1.6.0` which wasn't added as built-in library. This PR seems incomplete, but I might fix it when I've got the time.

JavkhlanK:
This PR did make my test project using a YouTube Player View work after pressing on Run though! Nice!

JavkhlanK:
wha

PR #861, opened 2023-04-29
fix: Images not loading in about screen in old android versions

HAKANKOKCU:
I fixed an issue happening in old android (before android 7), the images at about screen had shandows but not images. 

# Before:
Screenshot_2023-04-29-15-39-17

# After:
Screenshot_2023-04-29-15-27-08


TechnicalStudioDeveloper:
I already thought about this but thought I would not work.

PR #884, opened 2023-06-12
Pin Gradle actions to specific versions

ArjunaKumarMohanta:


JavkhlanK:
what's the point in pinning a specific version? doesn't that just force more frequent updates since automatic ones from e.g. 2.4.2 to 2.4.3 are disabled?

PranavPurwar:
why do they enquire you about sketchware pro buildscript

JavkhlanK:
> I consistently receive inquiries from individuals stating that the build script is susceptible to x/y/z vulnerabilities, even though the 'v2' tag has been redirected to a non-vulnerable version of these dependencies. This is why I make these changes.

just tell them that it's automatically been "redirected" to a non-vulnerable version then?

PranavPurwar:
we dont use configuration cache

JavkhlanK:
Just ignore that, the repo's already using v2:  
image

PR #886, opened 2023-06-12
Feat x Refactor

elfilibusterismo:
Added some class from a.a.a to com.besome.sketch...

Edited PaletteSelector to use the added OnBlockCategorySelectListener, LayoutUtil, and PaletteSelectorItem.

Edited LogicEditorActivity to use OnBlockCategorySelectListener instead of a.a.a.Vs


PranavPurwar:
could you get rid of those unnecessary new lines? thanks

TheRemakerMan:
> could you get rid of those unnecessary new lines? thanks

I think these empty lines are automatically added due to some poor quality code editor.

JavkhlanK:
what's the point in copying already existing classes with only readable names? at least I'd have started with decompiling e.g. `a.a.a.nA` and cleaning up its code (but not touching the method signatures if used in still undecompiled classes!)

JavkhlanK:
also, I'll be honest, why are you practically wasting your time in the GitHub.com web editor? doesn't seem like you tested out your changes and they neither are formatted properly.

elfilibusterismo:
Sorry my bad...

PR #887, opened 2023-06-13
refactor: Refactored class Dp to ProjectCompiler

TechnicalStudioDeveloper:


PranavPurwar:
variable names can be changed to just `compiler` instead of `projectCompiler`

JavkhlanK:
I think `ProjectBuilder` would be a better name. The class doesn't do the actual compiling (converting code into something else).

btw a better commit message would've been *renamed* Dp to ProjectCompiler

PR #896, opened 2023-06-27
feat: improve kotlin compiler speed and upgrade compiler

PranavPurwar:


README.md
```diff
@@ -1,6 +1,11 @@
-<p align="center">
-  <img src="assets/Sketchware-Pro.png" />
+<p>
+  <img src="assets/Sketchware-Pro.png"  alt="Sketchware Pro logo"/>
 </p>
+<style>
+p {
+    text-align: center;
+}
+</style>
```
JavkhlanK, line 8: this breaks the centered image [even though the previous approach is the only one that works](https://stackoverflow.com/a/62383408)

kotlinc/src/main/java/org/jetbrains/kotlin/cli/jvm/compiler/jarfs/FastJarFileSystem.kt
```diff
@@ -0,0 +1,131 @@
+/*
+ * This file is part of Cosmic IDE.
```
JavkhlanK, line 2: didn't you remove the original license and replace it with Cosmic IDE's?

JavkhlanK (changes requested):

kotlinc/src/main/java/org/jetbrains/kotlin/cli/jvm/compiler/jarfs/FastJarFileSystem.kt
PranavPurwar: ~~copied the module from cosmic ide, it was auto edited/added by AS~~

PranavPurwar: restored the original license header on files that haven't been modified from source

README.md
PranavPurwar: fixed

PR #898, opened 2023-07-04
bug fix when lib not download

PranavPurwar:
reported on telegram

### DIFF
--- a/app/src/main/java/mod/pranav/dependency/resolver/DependencyResolver.kt
+++ b/app/src/main/java/mod/pranav/dependency/resolver/DependencyResolver.kt
@@ -147,8 +147,10 @@ class DependencyResolver(
                 callback.onDependencyResolveFailed(e)
                 return@forEach
             }
-            callback.log("Cannot download ${artifact.toStr()}")
-            if (path.toFile().exists().not()) return@forEach
+            if (path.toFile().exists().not()) {
+                callback.log("Cannot download ${artifact.toStr()}")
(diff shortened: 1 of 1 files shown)


app/src/main/java/mod/pranav/dependency/resolver/DependencyResolver.kt
```diff
@@ -147,6 +147,8 @@ class DependencyResolver(
                 callback.onDependencyResolveFailed(e)
                 return@forEach
             }
+            callback.log("Cannot download ${artifact.toStr()}")
```
JavkhlanK, line 150: is the artifact really not downloadable when it reaches that statement?

PranavPurwar: I guess so, since the artifact failed earlier, it's possible that user configuration is incorrect or the system refused the request, or the error would have been logged and shown

JavkhlanK: wdym failed earlier? what I meant is are you sure you added that log statement in the right place since it looks like that'll be called every time a valid dependency is resolved

PranavPurwar: yes

JavkhlanK: epic

PR #904, opened 2023-07-10
chore: Added ability to search for icons

elfilibusterismo:


JavkhlanK:
1. wdym chore? it's a feature if anything, not something like a cleanup
2. what's up with the formatting??
3. did you create the file in CRLF? the whole repo is LF-only, please change that

PR #917, opened 2023-07-17
feat: Added badge in README file

DevKumar00001:


PranavPurwar:
For the CI badge, its no use since GitHub already shows that besides commit name

DevKumar00001:
> And other badges should link to the respective pages like for contributors and commits

I have not changed anything in old badge.

JavkhlanK:
i don't really see the need for that CI or repository size badge, but that downloads one seems neat

DevKumar00001:
> i don't really see the need for that CI or repository size badge, but that downloads one seems neat

Repo size badge is not bad.For first time when I cloned this repository I was unknown how much repository size is with `.git`  dir.

PR #942, opened 2023-08-06
feat: Enhanced Dependency Resolver (Speed)

CyberXCoder-6649:
By adding caching mechanism to avoid redundant loop of resolution currently it check if already dependency downloaded or resolved, with this resolution process is now faster than before

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -111,6 +111,8 @@ dependencies {
     implementation "com.android.tools:annotations:31.0.2"
     implementation "com.google.guava:guava:32.0.1-jre"
 
+    implementation "org.apache.maven:maven-artifact:3.8.1"
+    
     def editorGroupId = "io.github.Rosemoe.sora-editor"
     implementation platform("$editorGroupId:bom:0.21.1-f243159-SNAPSHOT")
     implementation "$editorGroupId:editor"
--- a/app/src/main/java/mod/pranav/dependency/resolver/DependencyResolver.kt
+++ b/app/src/main/java/mod/pranav/dependency/resolver/DependencyResolver.kt
@@ -74,7 +74,6 @@ class DependencyResolver(
         Environment.getExternalStorageDirectory().absolutePath,
         ".sketchware", "libs", "repositories.json"
     )
-
     init {
         if (Files.notExists(repositoriesJson)) {
             Files.createDirectories(repositoriesJson.parent)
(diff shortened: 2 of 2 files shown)


PranavPurwar:
This PR shall be closed (not merged) now.

JavkhlanK:
so if I understood this right, it tries to deduplicate dependencies. why should this not be merged, @PranavPurwar?

PranavPurwar:
> so if I understood this right, it tries to deduplicate dependencies. why should this not be merged, @PranavPurwar?

Duplication of dependencies according to their version is already taken care of as described in a review comment

CyberXCoder-6649:
> > so if I understood this right, it tries to deduplicate dependencies. why should this not be merged, @PranavPurwar?
> 
> Duplication of dependencies according to their version is already taken care of as described in a review comment

Ya but it was doing that at last, i have added caching mechanism, check if library is downloaded and skip, and compare version. (while resolving) by this it improves speed

PranavPurwar:
> > > so if I understood this right, it tries to deduplicate dependencies. why should this not be merged, @PranavPurwar?
> > 
> > 
> > Duplication of dependencies according to their version is already taken care of as described in a review comment
> 
> Ya but it was doing that at last, i have added caching mechanism, check if library is downloaded and skip, and compare version. (while resolving) by this it improves speed

Umm `resolvePOM()` does it and is called much earlier than when the new sorting takes place???

CyberXCoder-6649:
> > > > so if I understood this right, it tries to deduplicate dependencies. why should this not be merged, @PranavPurwar?
> > > 
> > > 
> > > Duplication of dependencies according to their version is already taken care of as described in a review comment
> > 
> > 
> > Ya but it was doing that at last, i have added caching mechanism, check if library is downloaded and skip, and compare version. (while resolving) by this it improves speed
> 
> Umm `resolvePOM()` does it and is called much earlier than when the new sorting takes place???

Could you explain

PranavPurwar:
> > > > > so if I understood this right, it tries to deduplicate dependencies. why should this not be merged, @PranavPurwar?
> > > > 
> > > > 
> > > > Duplication of dependencies according to their version is already taken care of as described in a review comment
> > > 
> > > 
> > > Ya but it was doing that at last, i have added caching mechanism, check if library is downloaded and skip, and compare version. (while resolving) by this it improves speed
> > 
> > 
> > Umm `resolvePOM()` does it and is called much earlier than when the new sorting takes place???
> 
> Could you explain? What u mean to say?

I mean that the thing you were saying it was doing last earlier, is actually done much earlier than when your changes do (check that resolvePOM() method is called in the beginning of the method)

PR #956, opened 2023-08-29
Fix moreblock collection preview in list item

elfilibusterismo:
Screenshot_2023-08-30-03-07-10-46_423e58f25c30a02453fa5d8c82fc37f2

### DIFF
--- a/app/src/main/java/com/besome/sketch/editor/manage/ManageCollectionActivity.java
+++ b/app/src/main/java/com/besome/sketch/editor/manage/ManageCollectionActivity.java
@@ -52,24 +52,19 @@ import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Iterator;
 
-import a.a.a.FB;
 import a.a.a.Mp;
 import a.a.a.Np;
 import a.a.a.Op;
 import a.a.a.Pp;
 import a.a.a.Qp;
 import a.a.a.Rp;
-import a.a.a.Rs;
-import a.a.a.Ts;
 import a.a.a.bB;
-import a.a.a.kq;
 import a.a.a.mB;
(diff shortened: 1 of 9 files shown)


app/src/main/java/mod/hey/studios/moreblock/importer/MoreblockImporterDialog.java
JavkhlanK, line 1: this commit does too much at once imo. could you split it up into multiple? for example one which makes the class extend `a.a.a.aB`, another making the class use `a.a.a.Rs`, and finally one making the dialog use `RecyclerView` instead of `ListView` (though I'm not really sure what the point of the RecyclerView one really is)

app/src/main/res/layout/manage_collection_popup_import_more_block_list_item.xml
```diff
@@ -47,10 +47,7 @@
                 <FrameLayout
                     android:id="@+id/block_area"
                     android:layout_width="wrap_content"
-                    android:layout_height="match_parent">
-
-                    <include layout="@layout/block_customview_spec" />
-                </FrameLayout>
+                    android:layout_height="match_parent"/>
```
JavkhlanK, line 51: this should be part of the commit which makes the dialog use `a.a.a.Rs`

app/src/main/java/com/besome/sketch/editor/manage/ManageCollectionActivity.java
JavkhlanK, line 1: i think the commit message should be something like "Show correct MoreBlock return type in …", or at least without that `a` since that's not proper English

JavkhlanK (changes requested):
overall another nice contribution! there's just a few improvements needed.

JavkhlanK:
i forgot the review stuff before merging lol

PR #961, opened 2023-09-04
Handle built in util classes

elfilibusterismo:
When SketchwareUtil and FileUtil blocks not used, it will not generate the source code.

### DIFF
--- a/app/src/main/java/mod/elfilibustero/sketch/lib/handle/BuiltInUtilClassHandler.java
+++ b/app/src/main/java/mod/elfilibustero/sketch/lib/handle/BuiltInUtilClassHandler.java
@@ -1,0 +1,90 @@
+package mod.elfilibustero.sketch.lib.handle;
+
+import a.a.a.eC;
+import a.a.a.hC;
+
+import com.besome.sketch.beans.ComponentBean;
+
+public class BuiltInUtilClassHandler {
+
+    /**
+     * Check if any SketchwareUtil block is used in the project.
+     */
+    public static boolean isSketchwareUtilBlockUsed(hC projectFileManager, eC projectDataManager) {
+        for (var activity : projectFileManager.b()) {
+            for (var entry : projectDataManager.b(activity.getJavaName()).entrySet()) {
+                for (var blockBean : entry.getValue()) {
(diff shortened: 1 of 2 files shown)


hasrat-ali (changes requested):
What if they use those methods in custom Java classes?

JavkhlanK:
it should be an opt-in feature if anything imo

elfilibusterismo:
> What if they use those methods in custom Java classes?

I've think of that too. Also in addSourceDirectly Block.

hasrat-ali:
Then why did you done this? 😂

elfilibusterismo:
> Then why did you done this? 😂

Cause what if you don't use that classes? It still generates even though you don't use that. 

Maybe what jbk said. Should have an option.

JavkhlanK:
breaking compilation should never really happen though

app/src/main/java/mod/elfilibustero/sketch/lib/handle/BuiltInUtilClassHandler.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        
+        return false;
+    }
+
+    /**
+     * Check if specific SketchwareUtil block is present.
+     */
+    public static boolean isSketchwareUtilPresent(String opCode) {
+        String[] blocks = {
```
PranavPurwar, line 31: this would create util class even if you use some custom block with this name, i think

JavkhlanK: wouldn't that be the fault of the Custom Block creator since they used an already existing Block opcode

PranavPurwar: i dont think they'd know that much about internals tbh

JavkhlanK: since that'd also override built-in Blocks' code and so on, this is nothing to worry about, it's just more damage from doing what shouldn't be done unless done carefully

PR #966, opened 2023-09-07
Binding support, and clean up unused resources

ilyassesalama:
I ensured everything works fine before opening this PR, but I still need others' reviews.


JavkhlanK (approved):
i checked for usages of all removed files by filename (occurrences of `layout.<filename>`, `@layout/<filename>`, and even just `<filename>` after manual filtering) and by resource ID in public-stable-ids.txt/public.xml and found no uses of them. Let's hope there's no massive usage of `Resources#getIdentifier(String, String, String)`

also did a quick runtime test without crashes

CyberXCoder-6649:
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:processMinApi21DebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.res.LinkApplicationAndroidResourcesTask$TaskAction
   > Android resource linking failed
     public-stable-ids.txt: error: failed reading stable ID file.


* Try:

When building Sketchware-Pro (assembleDebug)

JavkhlanK:
did you try building on a Windows host? that's untested by us, try using WSL and in a Linux filesystem if you are

CyberXCoder-6649:
> did you try building on a Windows host? that's untested by us, try using WSL and in a Linux filesystem if you are

Oh i tried Using AndroidIDE i will try with any other way

PR #968, opened 2023-09-07
Material redesign

PranavPurwar:
Add material you for ManageLibraryActivity, DesignDrawer, and some improvements in DesignActivity

### DIFF
--- a/.idea/intellij-javadocs-4.0.1.xml
+++ b/.idea/intellij-javadocs-4.0.1.xml
@@ -1,0 +1,204 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JavaDocConfiguration">
+    <GENERAL>
+      <MODE>UPDATE</MODE>
+      <OVERRIDDEN_METHODS>false</OVERRIDDEN_METHODS>
+      <SPLITTED_CLASS_NAME>true</SPLITTED_CLASS_NAME>
+      <LEVELS>
+        <LEVEL>METHOD</LEVEL>
+        <LEVEL>TYPE</LEVEL>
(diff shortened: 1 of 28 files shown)


.idea/intellij-javadocs-4.0.1.xml
ilyassesalama, line 1: Why you didn't exclude these files? I thought they're ignored.

JavkhlanK:
merged most of this in manually, but I think the rest in like ManageLibraryActivity or the compile log screen will be done when those activities will be fully made Material You

PranavPurwar:
> merged most of this in manually, but I think the rest in like ManageLibraryActivity or the compile log screen will be done when those activities will be fully made Material You

I thought I made ManageLibraryActivity material you in this pr??? the sub sections are different activities

JavkhlanK:
I said most. The activity will probably get a redesign to be like Local library Manager anyway, so those few M3 adjustments won't be needed anyway then.

app/src/main/java/com/besome/sketch/editor/view/ViewEditor.java
```diff
@@ -337,7 +337,6 @@ public boolean onTouch(View view, MotionEvent motionEvent) {
                     boolean z = false;
                     for (int i3 = 0; i3 < uyVar.getData().size(); i3++) {
                         ViewBean viewBean = uyVar.getData().get(i3);
-                        if (c(viewBean)) {
```
hasrat-ali, line 340: @PranavPurwar explanation?

JavkhlanK: look at the rest of the commit, the entire method used to do nothing except return `true`

PR #970, opened 2023-09-08
Fix some duplicate stuff

elfilibusterismo:


JavkhlanK:
Something tells me this shouldn't be added. It allows users to slack off and create duplicate imports. And even if, who cares, projects will still compile with or without these patches.

JavkhlanK:
For now I'll only cherry-pick that 3rd commit.

09230799146:
Ibn for what?

PR #976, opened 2023-09-11
Fix custom component

elfilibusterismo:


app/src/main/java/a/a/a/mq.java
```diff
@@ -760,7 +760,7 @@ public static String e(String typeName) {
             case "Context" -> "Activity";
             case "ResString", "ResStyle", "ResColor", "ResArray", "ResDimen", "ResBool", "ResInteger", "ResAttr", "ResXml", "Color" ->
                     "int";
-            default -> typeName;
+            default -> ComponentsHandler.getVarName(typeName);
```
JavkhlanK, line 763: okay, why exactly is this needed? I feel like making the varName -> typeName change was a mistake. I don't want to mess around with types and stuff, let's just keep stuff as it is since it (used to) work.

JavkhlanK: this commit message prefix also makes no sense, it's not a feature, it's a fix if anything

elfilibusterismo: > okay, why exactly is this needed? I feel like making the varName -> typeName change was a mistake. I don't want to mess around with types and stuff, let's just keep stuff as it is since it (used to) work.

It is not a mistake, you can check the logic of FirebaseDB how it's done. Key typeName is the main key of component. In a.a.a.mq#e used typeName(FirebaseDB) to get variable type name(DatabaseReference).

elfilibusterismo: Also this fix the issue for https://github.com/Sketchware-Pro/Sketchware-Pro/discussions/978#discussioncomment-6984092

guhg:
I hope this is fixed soon and merged with material redesign

app/src/main/java/a/a/a/mq.java
elfilibusterismo: Example For FirebaseDB

name - Firebase DB (This will show when you add component also in blocks when you use %m.firebase) 
typeName - FirebaseDB 
buildClass - FirebaseDB
class - Component.FirebaseDB(typeName)
varName - DatabaseReference (If you remove this in a.a.a.mq#e, when you add Firebase DB component, source code will show private FirebaseDB instead of private DatabaseReference)

JavkhlanK: I meant that commit 2bb5fcf4211491257979c57de976ed1267e53795 might've been a mistake, since it apparently wasn't tested enough since this had to be created. I don't feel safe digging into Sketchware logics like this.

PR #977, opened 2023-09-12
Admob App Id Setting

elfilibusterismo:


app/src/main/java/com/besome/sketch/editor/manage/library/admob/AddAppIdStepView.java
```diff
@@ -0,0 +1,102 @@
...
+    private void showAddAppIdDialog() {
+        aB dialog = new aB((Activity) getContext());
+        dialog.b(xB.b().a(getContext(), R.string.design_library_admob_dialog_set_app_id));
+        dialog.a(R.drawable.add_96_blue);
+        ManageLibrarySettingAdmobAppIdAddBinding addBinding = ManageLibrarySettingAdmobAppIdAddBinding.inflate(dialog.getLayoutInflater());
+        EditText edAppId = addBinding.edAppId;
+        edAppId.setText(appId);
+        edAppId.setPrivateImeOptions("defaultInputmode=english;");
+        dialog.a(addBinding.getRoot());
+        dialog.b(xB.b().a(getContext(), R.string.common_word_add), v -> {
```
hasrat-ali, line 90: Here you should use Helper method

app/src/main/java/com/besome/sketch/editor/manage/library/admob/AddAppIdStepView.java
```diff
@@ -0,0 +1,102 @@
...
+            String id = edAppId.getText().toString();
+            if (!isEmpty(id)) {
+                setAppId(id);
+                dialog.dismiss();
+            } else {
+                edAppId.requestFocus();
+            }
+        });
+        dialog.a(xB.b().a(getContext(), R.string.common_word_cancel), Helper.getDialogDismissListener(dialog));
```
hasrat-ali, line 99: Here too

app/src/main/java/com/besome/sketch/editor/manage/library/admob/ManageAdmobActivity.java
```diff
@@ -201,6 +209,7 @@ public void onCreate(Bundle savedInstanceState) {
         A = new DB(getApplicationContext(), "P1");
         admobLibraryBean = getIntent().getParcelableExtra("admob");
         ((TextView) findViewById(R.id.tv_enable)).setText(Helper.getResString(R.string.design_library_settings_title_enabled));
+        ((TextView) findViewById(R.id.tv_title_app_id)).setText(Helper.getResString(R.string.design_library_admob_title_app_id));
```
hasrat-ali, line 212: Since these strings aren't translated so you shouldn't use `Helper#getResString(int)`.
You can use that string in XML.

hasrat-ali (changes requested):
I think you shouldn't completly remove `setAdmobAppId` block.

elfilibusterismo:
> I think you shouldn't completly remove `setAdmobAppId` block.

Removing setAdmobAppId block completely doesn't break the project. It'll just show a purple block. And Also jbk suggested it.

app/src/main/java/com/besome/sketch/editor/manage/library/admob/ManageAdmobActivity.java
elfilibusterismo: So what you mean is I'll just use setText(R.string.string_name)?

JavkhlanK:
I thought of completely removing the Block since it won't be needed anymore. I had automatic checks like for the AdMob app ID when about to build a project in mind. That check would see if the app ID was set in AdMob settings and if not, prompt the user to set it.

We might as well add automatic removal of setAdmobAppId Blocks to that.

PR #1084, opened 2024-02-11
Some fixes

remaker17:
Some fixes at the request of contributors.


remaker17:
@coderabbitai review

hasrat-ali:
Okey, all looks good to me.

ilyassesalama:
The bottom sheet still shows even though there are no logs shown. Is this intentional? if yes, then it shouldn't.

PR #1087, opened 2024-02-12
Update SketchLogger.java

SENPAi-03:
Replace thread :)


SENPAi-03:
Improving

JavkhlanK:
And for what reason? Also this won't work with dx / Java 1.7 projects.

SENPAi-03:
Sometimes the app breaks down, also we can replace lambda syntax

JavkhlanK:
Sometimes? When? Breaks down? In what exact way?

PR #1129, opened 2024-05-19
MaterialAlertDialog animation in aB dialog

aquilesTrindade:


remaker17:
I'm sure alert dialog already has M3 animation.

aquilesTrindade:
> Tenho certeza de que a caixa de diálogo de alerta já possui animação M3.

no, there isn't

ilyassesalama:
Matarial3 already has dialog fade in/out animation by default. Thanks for the contribution though.

remaker17:
> Matarial3 already has dialog fade in/out animation by default. Thanks for the contribution though.

I don't think so, because we are using custom dialog (android.app.Dialog).

ilyassesalama:
I tested with and without his change before closing this PR. There was no difference.

aquilesTrindade:
😂 unbelievable

PR #1133, opened 2024-05-27
Improve ViewBinding handling in CommonViewBindingAdapter

remaker17:
See title.


hasrat-ali:
Well, eeehhh
I did this so that it would be convenient when we have to use multiple `viewType.`

hasrat-ali:
For example: EmptyView, LoadingView etc.

remaker17:
Oh, I almost forgot about viewTypes.
I'll close it then.

PR #1141, opened 2024-05-30
Update DependencyResolver

PranavPurwar:
Greatly improved speed as well as many bug fixes

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -139,7 +139,7 @@ dependencies {
     implementation "com.squareup.okhttp3:okhttp:4.10.0"
     //noinspection GradleDependency
     implementation "com.squareup.okio:okio:3.1.0"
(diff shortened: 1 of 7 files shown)


JavkhlanK:
Why does the commit named "Update DependencyResolver to v1.1.1" also bump the compile SDK version, other dependencies' versions, update the Gradle version, and add more callbacks to the downloader? Please use Git correctly.

PranavPurwar:
> Why does the commit named "Update DependencyResolver to v1.1.1" also bump the compile SDK version, other dependencies' versions, update the Gradle version, and add more callbacks to the downloader? Please use Git correctly.

"More callbacks" are actually needed to adapt to the new structure of the library, so that's why

PranavPurwar:
Changes of this PR have been included in #1162

PR #1144, opened 2024-05-31
Update artifacts version

remaker17:
It fixes:
Screenshot


JavkhlanK:
Why not just e.g. `@v4`? Doesn't that also automatically use newer v4 versions when those are released?

remaker17:
Looks good to me!

PR #1143, opened 2024-05-31
Add copy error to clipboard feature in `CollectErrorActivity`

PranavPurwar:
Its pretty annoying to read incomplete error messages from screenshots

PR #1148, opened 2024-06-01
fix,feat: Improve DebugActivity

### LINKED ISSUE #1140: When App Crashes, Debug Screen no longer shows with error
If my app crashes, Debug activity doesn't come up anymore so I don't know what the error message is. And I can't use logcat because I don't have root.

khaled-0:
fixes #1140 
DebugActivity now launched every time using intent because,
AlarmManager is unreliable

Also changed the crash message view from dialog to a bi directional scroll
Profit: formatted crash logs

Screenshot from 2024-06-01 13-32-45

### DIFF
--- a/app/src/main/assets/debug/DebugActivity.java
+++ b/app/src/main/assets/debug/DebugActivity.java
@@ -39,7 +39,7 @@ public class DebugActivity extends Activity {
         String madeErrorMessage = "";
 
         if (intent != null) {
-            errorMessage = intent.getStringExtra("stacktrace");
+            errorMessage = intent.getStringExtra("error");
 
             String[] split = errorMessage.split("\n");
             //errorMessage = split[0];
@@ -65,20 +65,17 @@ public class DebugActivity extends Activity {
         }
 
         setTitle(getTitle() + " Crashed");
-        setContentView(contentView(this, madeErrorMessage));
-    }
 
-    View contentView(Context context, String errorMessage) {
-        TextView error = new TextView(context);
-        error.setText(errorMessage);
+        TextView error = new TextView(this);
+        error.setText(madeErrorMessage);
         error.setTextIsSelectable(true);
 
-        HorizontalScrollView hscroll = new HorizontalScrollView(context);
-        ScrollView vscroll = new ScrollView(context);
+        HorizontalScrollView hscroll = new HorizontalScrollView(this);
+        ScrollView vscroll = new ScrollView(this);
 
         hscroll.addView(vscroll);
         vscroll.addView(error);
 
-        return hscroll;
+        setContentView(hscroll);
     }
 }
--- a/app/src/main/assets/debug/DebugActivity.java
+++ b/app/src/main/assets/debug/DebugActivity.java
@@ -6,7 +6,11 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
 import android.util.Log;
+import android.widget.HorizontalScrollView;
+import android.widget.ScrollView;
 import android.widget.TextView;
+import android.content.Context;
+import android.view.View;
 
 public class DebugActivity extends Activity {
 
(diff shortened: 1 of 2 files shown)


app/src/main/assets/debug/SketchApplication.java
```diff
@@ -20,33 +17,19 @@ public static Context getContext() {
...
-
-                    uncaughtExceptionHandler.uncaughtException(thread, throwable);
-                }
-            });
+                new Thread.UncaughtExceptionHandler() {
+                    @Override
+                    public void uncaughtException(Thread thread, Throwable throwable) {
+                        Intent intent = new Intent(getApplicationContext(), DebugActivity.class);
+                        intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+                        intent.putExtra("stacktrace", Log.getStackTraceString(throwable));
```
JavkhlanK, line 27: This will break existing projects with a custom crash handler.

app/src/main/assets/debug/DebugActivity.java
```diff
@@ -60,17 +64,21 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                    }
-                })
-                .create();
-        dialog.show();
-        ((TextView) dialog.findViewById(android.R.id.message)).setTextIsSelectable(true);
+        setTitle(getTitle() + " Crashed");
+        setContentView(contentView(this, madeErrorMessage));
+    }
+
+    View contentView(Context context, String errorMessage) {
```
JavkhlanK, line 71: Why a separate method? Why pass over `Context` in a non-static method?

JavkhlanK (changes requested):

app/src/main/assets/debug/DebugActivity.java
khaled-0: ??? The method just returns the view, what's that got to do with non static

I put it in a method without any specific reason. just keeping the view buildings out of onCreate

app/src/main/assets/debug/SketchApplication.java
khaled-0: explain

JavkhlanK: They rely on the extra being named `error`.

app/src/main/assets/debug/DebugActivity.java
JavkhlanK: It'd just be cleaner without an unnecessary extra method. The parameter is redundant too, since `this` can be used (it's not a static method).

JavkhlanK:
Works great. What's suboptimal is that you can't easily take a screenshot of the entire stack trace anymore. I'm 100 % sure most users would not care and then start sending cut-off ones, making every helper's life harder. I think the horizontal scroll feature should be removed.

khaled-0:
> Works great. What's suboptimal is that you can't easily take a screenshot of the entire stack trace anymore. I'm 100 % sure most users would not care and then start sending cut-off ones, making every helper's life harder. I think the horizontal scroll feature should be removed.

Or how about a copy button?

JavkhlanK:
They'll probably still ignore that.

khaled-0:
I dunno, go ahead and remove it then i guess

guhg:
I think just adding a copy button above or below the error code with be good enough

hasrat-ali:
Can't we just add inbuilt screenshot button?

JavkhlanK:
> I think just adding a copy button above or below the error code with be good enough

There's a copy button in Sketchware Pro's crash screen. Not a lot of people used it afair. At least the messages weren't cut off in the dialog.

PranavPurwar:
> > I think just adding a copy button above or below the error code with be good enough
> 
> There's a copy button in Sketchware Pro's crash screen. Not a lot of people used it afair. At least the messages weren't cut off in the dialog.

There isn't, yet.

JavkhlanK:
wait maybe I misremembered lmao

PranavPurwar:
yeah, it was never there so I made #1143

PR #1152, opened 2024-06-04
fix, style: Project deletion is now done via a dialog and the dialogs in the Android Manifest Manager event have been changed to aB dialog.

Unepix:
I'm not well versed in Android Studio and more precisely in things like Git, so I'm sorry if I made a mistake.

I did some work and here are the parts I changed for now;

- I disabled the Delete Project buttons and added aB dialog instead (I didn't remove the keys completely, just disabled them)
- When you cancel the build while the project is being created, the build dialog is closed and the Cancelling build information is added to the run button so you don't wait for the project to be canceled. (This worked for me but I'm not sure if it's stable)
- I converted Launcher Activity, Add Activity, Remove Activity dialogs in Android Manifest Manager page to aB dialog.


ilyassesalama:
Why did you close this pull request?

Unepix:
> Bu çekme isteğini neden kapattınız?



> Bu çekme isteğini neden kapattınız?

Dude, I accidentally shared the same thing twice, so I closed it, but I hope there was no problem.

PR #1177, opened 2024-06-29
feat: upgrade kotlinc to v2.0.0

PranavPurwar:
Upgrades kotlin compiler to v2.0.0. Also I had already published the port on Jitpack so I guess there's no point of having them increase repository size here.


remaker17:
Thanks for you contributions. We hope that you will continue to support Sketchware Pro in the future. Every contribution is very valuable to us.

PR #1182, opened 2024-07-01
feat : Improvements for SketchApplication.java

Yamenher:
Improvements for better Unhandled errors catching on newer android versions as `AlarmManager` Component is not responsive on recent android versions, I think this code belongs to @JavkhlanK originally :/.


remaker17:
Duplicate of #1148.

PR #1196, opened 2024-07-06
Fix: Non-cancelable dependency download dialog

aquilesTrindade:


hasrat-ali:
I think this should not be done because `material-redesign` branch will be merged soon. And there will be conflict between them.

aquilesTrindade:
> Acho que isso não deve ser feito porque `material-redesign`o branch será mesclado em breve. E haverá conflito entre eles.

hum, ok

PR #1226, opened 2024-08-23
fix view colors on BuildSettingsDialog

antsummer:
fix view colors on BuildSettingsDialog and add a class to "facilitate" the use of TextInputLayout & TextInputEditText  in java or kt class 

See:

Screenshot_20240823-005223_Sketchware Pro
Screenshot_20240823-005241_Sketchware Pro

### DIFF
--- a/app/src/main/java/mod/hey/studios/build/BuildSettingsDialog.java
+++ b/app/src/main/java/mod/hey/studios/build/BuildSettingsDialog.java
@@ -151,7 +151,6 @@ public class BuildSettingsDialog {
 
     private EditText addInputPref(String key, String defaultValue, String hint, int inputType, LinearLayout addTo) {
         TextInputLayout textInputLayout = new TextInputLayout(activity);
-
         LinearLayout.LayoutParams layoutParams = new LinearLayout.LayoutParams(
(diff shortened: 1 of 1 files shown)


ilyassesalama:
Why did you create a whole new `SKTextField` class instead of using `TextInputLayout` in XML?

antsummer:
@ilyassesalama because the dialog uses java and not xml, and even with hasrat did, the edittext did not look like the one in the photo sent

ilyassesalama:
> @ilyassesalama because the dialog uses java and not xml, and even with hasrat did, the edittext did not look like the one in the photo sent

We're trying to migrate UI stuff from Java to XML as much as possible to separate design to an easy, customizable, and reusable way. If the screen you're editing does not have an XML, you can create one. Though, your changes are nice.

antsummer:
> > @ilyassesalama because the dialog uses java and not xml, and even with hasrat did, the edittext did not look like the one in the photo sent
> 
> We're trying to migrate UI stuff from Java to XML as much as possible to separate design to an easy, customizable, and reusable way. If the screen you're editing does not have an XML, you can create one. Though, your changes are nice.

yeah I had thought about it, but when I did it it was too late, I could have waited until another day, but ok

PR #1252, opened 2024-09-15
feat: enable predictive gestures.

trindadedev13:
preview:

https://github.com/user-attachments/assets/4c20b3b7-37ec-4d41-8d7f-419adad4e6ee

https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture?hl=pt-br

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -23,6 +23,7 @@
         android:largeHeap="true"
         android:requestLegacyExternalStorage="true"
         android:theme="@style/Theme.SketchwarePro"
+        android:enableOnBackInvokedCallback="true"
         android:usesCleartextTraffic="true">
 
         <provider


ilyassesalama:
Even though I tested this and it seems pretty good, I don't think this feature is 100% stable yet. Sometimes, It is glitchy when going back quickly and may make some activities blink. However, this gesture runs smoothly when you swipe and hold to see the animation.

This is my experience on a Samsung device, I don't know about other devices. We'll wait for more feedback from the others before merging this.

trindadedev13:
@ilyassesalama Yeah, the gesture of haste really isn't very interesting..., I tested it on 2 cell phones (moto g34 & Moto g53) and the result was the same:

https://github.com/user-attachments/assets/8d386136-0e2e-4823-b2ba-cdafe4bcba08

ilyassesalama:
In your video, look at the last moment of it. Did you notice how the activity glitched?

ilyassesalama:
it seems like the same issue occurred on your device as well. Notice how the activity glitched at the end of the video. That's the issue I was talking about.

trindadedev13:
@ilyassesalama yes, I hadn't tested this before

JavkhlanK:
This requires that every single activity and fragment using custom onBackPressed behavior must be changed to work with this. This is far from complete.

PR #1256, opened 2024-09-20
feat: Material progress indicator in projects list

skinvent:


trindadedev13:
In my opinion, it didn't turn out very well, especially when the animation happens quickly.

ilyassesalama:
> In my opinion, it didn't turn out very well, especially when the animation happens quickly.

Yes, that's why I self-assigned this pull request to investigate better approaches before merging it.

trindadedev13:
> > In my opinion, it didn't turn out very well, especially when the animation happens quickly.
> 
> Yes, that's why I self-assigned this pull request to investigate better approaches before merging it.

and what is the difference between assigning it or not? like, what does that change?

app/src/main/res/layout/progress.xml
```diff
@@ -6,21 +6,9 @@
...
 
-    <com.airbnb.lottie.LottieAnimationView
-        android:id="@+id/anim_sketchware"
-        android:layout_width="82dp"
-        android:layout_height="82dp"
-        android:layout_gravity="center"
-        app:lottie_autoPlay="true"
-        app:lottie_fileName="anim/sketchware_logo.json"
-        app:lottie_loop="true" />
+<com.google.android.material.loadingindicator.LoadingIndicator
```
trindadedev13, line 9: When I tested this LoadingIndicator, I set its size to 100dp and it looked nice on my phone, so I think it would look nice here too.
this

skinvent:
Ngl it's good and some guys can't see cause they have Myopia and hyperopia or the both at the same time so when you do an exact value for a loading indicator can cause UI issues for some people who make screen too big so it's good and enough and please close this discussion

app/src/main/res/layout/progress.xml
skinvent: Ngl it's good and some guys can't see cause they have Myopia and hyperopia or the both at the same time so when you do an exact value for a loading indicator can cause UI issues for some people who make screen too big so it's good and enough and please close this discussion

ilyassesalama:
> Ngl it's good and some guys can't see cause they have Myopia and hyperopia or the both at the same time so when you do an exact value for a loading indicator can cause UI issues for some people who make screen too big so it's good and enough and please close this discussion

A reply like was so unnecessary.

trindadedev13:
> Ngl it's good and some guys can't see cause they have Myopia and hyperopia or the both at the same time so when you do an exact value for a loading indicator can cause UI issues for some people who make screen too big so it's good and enough and please close this discussion

anyway it was changed to 100 and on my phone it was fine.

PR #1261, opened 2024-09-23
feat: Better UI on BlocksManager

trindadedev13:


ilyassesalama:
I guess I'll keep this pull request open for a while until you or I make serious changes to Block Manager. It seems like it needs many UI improvements.

PR #1262, opened 2024-09-23
Minimum SDK 24

subhash16:


JavkhlanK:
any benefit? i don't see any

subhash16:
> any benefit? i don't see any

Target SDK 34 so minimum SDK should 24 coz 21 is too low

JavkhlanK:
that's not a benefit. the target SDK is not related to the minimum, you can change both, regardless of the other. you need an actual benefit, like "library x only supports minSdk 24 and it's used in 70 % of projects" (but that doesn't exist)

subhash16:
> that's not a benefit. the target SDK is not related to the minimum, you can change both, regardless of the other. you need an actual benefit, like "library x only supports minSdk 24 and it's used in 70 % of projects" (but that doesn't exist)

Ok

PR #1267, opened 2024-09-28
feat: use MaterialButtonToggleGroup instead ChipGroup

trindadedev13:
* feat: use MaterialButtonToggleGroup instead ChipGroup

* style: add top margin as 10dp

a basic change


trindadedev13:
Screenshot_20240927-234536

ilyassesalama:
In my opinion, this entire screen should be removed. It doesn't make any sense to have an entire activity just for changing the app's theme. Many settings have to merged into one universal screen.

ilyassesalama:
Why don't add it for these instead?

Screenshot_20240930_162304_Sketchware Pro

trindadedev13:
Screenshot_20240930-170531
like this? I just think I should decrease the spacing

ilyassesalama:
Great, you only need to wrap `MaterialButtonToggleGroup` inside `HorizontalScrollView` to show all options since we should add bottom sheet fragment option as well. You may want to open a new pull request with these changes.

trindadedev13:
@ilyassesalama I'm not done yet, but when I'm done I'll open the other PR, I just changed the UI [13a7e47](https://github.com/trindadedev13/Sketchware-Pro/commit/13a7e47083aa47864942c662cf1f7c2ede8374a4)

PR #1290, opened 2024-10-15
feat, style: Suggestions for inject, remove colors definition of AutoCompleteTextView style

trindadedev13:


trindadedev13:
I don't know if AutoCompleteTextView looks good in "Custom Attributes" or inject, what do you think?

PR #1292, opened 2024-10-16
fix: Crash when use res values in ```Custom Attributes```

trindadedev13:
this does not add support for res values, but causes sketchware to crash, I even started trying to make support for res values, but stopped


trindadedev13:
> Can you tell me how can I reproduce this crash?

app:cornerRadius="@dimen/some_dimen"

use this on a button, and you will crash.

trindadedev13:
> > Can you tell me how can I reproduce this crash?
> 
> app:cornerRadius="@dimen/some_dimen"
> 
> use this on a button, and you will crash.

and no, it doesn't only happen with this code, it happens with any code that is being used in the preview that uses something from res instead of the direct value

elfilibusterismo:
Modifying PropertiesUtil#resolveSize to handle that would be easier

trindadedev13:
> Modifying PropertiesUtil#resolveSize to handle that would be easier

Yesss really, I hadn't thought of that

trindadedev13:
closing because I probably won't finish

ilyassesalama:
Pensive

Mester-Dark:
Oh Sorry ...

trindadedev13:
what

PR #1293, opened 2024-10-16
New Logic Palette

elfilibusterismo:
Replaced RecyclerView(CategoryList) with NavigationRailView


trindadedev13:
Screenshot_20241016-190755
very good, I just think it could be centered, or all the way at the top

injector-origin:
> Screenshot_20241016-190755 beautiful 😻, I just think it could be centered, or all the way at the top

Screenshot_20241017-063457
Already fixed

Yamenher:
I think you should use some padding between the fab and navigation rail view it's snot very good like that

PR #1297, opened 2024-10-17
feat: Search in blocks palettes

Edward-NewGate-KOC:


ilyassesalama:
Great addition! However, next time, please try to write more efficient code. I reviewed your changes, and it seems that almost 60% of the code could be removed. Additionally, we try to avoid code repetition, so be sure to use our existing classes, such as `aB` for showing dialogs, instead of `AlertDialog`.

Also, I removed the `startsWith` and `contains` radio buttons, leaving only `contains` as the default, as they serve the same purpose. This simplification will make the search process easier for users.

Check my commit for best practices to write code for Sketchware Pro codebase.

CliipCode:
رائع يا اخي الياس اود ان اشارككم في عملكم

PR #1303, opened 2024-10-19
Block manager preview

elfilibusterismo:
Use Rs class instead of 9 patch for block preview

Screenshot_20241019-211207


ilyassesalama:
Outstanding work like usual. Thanks for the great contribution.

PR #1311, opened 2024-10-22
Improve Project preview and add transitions

PranavPurwar:
Also migrates `SingleCopyAsyncTask` to coroutines


trindadedev13:
hey @PranavPurwar why didn't you use data class in the project template is it bad or something?

Issue #1224, opened 2024-08-15
InflateException: Missing `layout_width` Attribute in XML Layout Despite Proper Configuration


RoyTheReal:
**Issue Type**: Urgent bug

**Environment**:
- **App Name**: Waveline (My app)
- **Sketchware Pro version**: Version v6.4.0-SNAPSHOT-135ad63-minApi26
- **Package Name**: `com.fooks.technology.waveline`
- **Activity**: `LoginActivity in my app`
- **OS Version**: Android 14

**Summary**:
The app crashes with an `InflateException` stating that a `layout_width` attribute is missing, even though the XML layout file appears to be correctly configured with all necessary attributes.

**Error Message**:
```java
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.fooks.technology.waveline/com.fooks.technology.waveline.LoginActivity}: android.view.InflateException: Binary XML file line #1 in com.fooks.technology.waveline:layout/login: Binary XML file line #1: You must supply a layout_width attribute., theme={InheritanceMap=[...]}
Caused by: java.lang.UnsupportedOperationException: Binary XML file line #1: You must supply a layout_width attribute.
```
**Full Error Message**:
[...]

Issue #1118, opened 2024-04-14
AppCompat injection Manager bugs 


TanTai97o:
**Describe the bug**

https://github.com/Sketchware-Pro/Sketchware-Pro/assets/107351062/60b913b0-ded1-4e82-b773-759b6afee9ee

Device and Sketchware Pro-related info:
 - Device: SamSungA51 
 - OS:  Android 13
 - Sketchware Pro version: v6.4.0-SNAPSHOT-7be66bf-minApi26

PR #1315, opened 2024-10-23
Fix appcompat attributes deletion

elfilibusterismo:
This will fix the issue https://github.com/Sketchware-Pro/Sketchware-Pro/issues/1224


elfilibusterismo:
Same with issue https://github.com/Sketchware-Pro/Sketchware-Pro/issues/1118

PR #1317, opened 2024-10-23
Resolve issue with drag and drop

elfilibusterismo:
This pr is supposed to solve this issue https://github.com/Sketchware-Pro/Sketchware-Pro/issues/1219

However, this changes need more testing.

### DIFF
--- a/app/src/main/java/com/besome/sketch/editor/view/ViewPane.java
+++ b/app/src/main/java/com/besome/sketch/editor/view/ViewPane.java
@@ -646,107 +646,78 @@ public class ViewPane extends RelativeLayout {
         }
         return result;
     }
-
+    
+    private Rect getRectFor(View view) {
+        var rect = new Rect();
+        view.getGlobalVisibleRect(rect);
+        int scaledWidth = (int) (view.getWidth() * getScaleX());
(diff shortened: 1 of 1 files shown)


PranavPurwar:
"needs more testing"
Ilyasse: lets merge it rn

ilyassesalama:
> "needs more testing" Ilyasse: lets merge it rn

IMG_0011

PranavPurwar:
It works for some doesn't mean it will work for everyone, thats why PRs are thoroughly tested before merging rather than on user feedback alone.

elfilibusterismo:
> It works for some doesn't mean it will work for everyone, thats why PRs are thoroughly tested before merging rather than on user feedback alone.

Actually the "some" you're talking about is the ones who experiencing the issue

PranavPurwar:
> > It works for some doesn't mean it will work for everyone, thats why PRs are thoroughly tested before merging rather than on user feedback alone.
> 
> Actually the "some" you're talking about is the ones who experiencing the issue

Yes, but what if it causes some unexpected behaviour? What if it breaks backwards compatibility? That's why I think its necessary to thoroughly test it. especially because it deals with legacy code.

Issue #1095, opened 2024-02-23
Event Component bugs


TanTai97o:
**Describe the bug**

https://github.com/Sketchware-Pro/Sketchware-Pro/assets/107351062/dc4dc10b-65a7-4ae2-a4e8-80eaa6640b11

This bug also exists in the material redesign version

Device and Sketchware Pro-related info:
 - Device: SamSungA51 
 - OS: Android 13
 - Sketchware Pro version: v6.4.0-SNAPSHOT-ccf646e-minApi26

Issue #1061, opened 2024-01-12
Error adding component event 


TanTai97o:
**Describe the bug**

java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.get(ArrayList.java:437)
	at java.util.Collections$UnmodifiableList.get(Collections.java:1356)
	at a.a.a.br$Adapter$ViewHolder$EventViewHolder.lambda$new$0(br.java:277)
	at a.a.a.br$Adapter$ViewHolder$EventViewHolder.$r8$lambda$1-x2wvEUHQ3fVbNUkTfokVn8NR4(Unknown Source:0)
	at a.a.a.br$Adapter$ViewHolder$EventViewHolder$$ExternalSyntheticLambda0.onClick(Unknown Source:2)
	at android.view.View.performClick(View.java:7881)
	at android.view.View.performClickInternal(View.java:7858)
	at android.view.View.-$$Nest$mperformClickInternal(Unknown Source:0)
	at android.view.View$PerformClick.run(View.java:30863)
	at android.os.Handler.handleCallback(Handler.java:942)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loopOnce(Looper.java:226)
	at android.os.Looper.loop(Looper.java:313)
	at android.app.ActivityThread.main(ActivityThread.java:8741)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)

Device and Sketchware Pro-related info:
 - Device: SamSungA51
 - OS: Android13
 - Sketchware Pro version: Version v6.4.0-SNAPSHOT-be51bbb-minApi26

JavkhlanK:
The only reason I asked you to use the template is to also get to know the steps to reproduce the crash. Why didn't you fill those sections out?

TanTai97o:
> The only reason I asked you to use the template is to also get to know the steps to reproduce the crash. Why didn't you fill those sections out?

@JavkhlanK 
Sorry about that
Just click here,sketchware is crashes
![Screenshot_20240112_193008_Sketchware Pro](https://github.com/Sketchware-Pro/Sketchware-Pro/assets/107351062/a22c104f-393b-4b3e-ba11-0f1b62eb5108)

PR #1319, opened 2024-10-24
fix: Crash when adding a new component in certain cases

elfilibusterismo:
This update might fix the crash when adding a new component event. I moved the code to their adapter, so it no longer depends on checking the type with instanceof. This should make the event more reliable.


hasrat-ali:
Is this a fix for #1061 and #1095

elfilibusterismo:
> Is this a fix for #1061 and #1095

Yes bro. This solves that issue

Yamenher:
@ilyassesalama please merge this

PR #1328, opened 2024-10-26
Feat : Color Editor + minor bug fixes

Mustiy0:
most of my color editor code is from strings editor


ilyassesalama:
I will keep this pull request open until the code is refactored, because I noticed that there are many repeated codes from strings editor activity. As well as there are some UI fixes need to be done.

PR #1339, opened 2024-10-29
feature: ViewBinding support

PranavPurwar:
Currently a WIP.
No option to switch it off yet.
Also decompiles and refactors `Fx` so there maybe unknown bugs/crashes although I didn't get any

For testing purposes only

PR #1347, opened 2024-11-01
Some Improvements

Edward-NewGate-KOC:


ilyassesalama:
Pranav already rewrote `Fx` class in #1339. I don't it's useful in this PR now.

Edward-NewGate-KOC:
> Pranav already rewrote `Fx` class in #1339. I don't think it's useful in this PR now.

The `Fx` class was indeed taken from `Pravav’s PR`, but I need it to generate the blocks code correctly ( it's better than using `builtInBlocks method` ). If we go back to the old method, the error I explained will return, and we’ll completely lose the quick access feature to the StringEditor from the blocks palettes.

Edward-NewGate-KOC:
> Pranav already rewrote `Fx` class in #1339. I don't think it's useful in this PR now.

bro, please check the last commit to know exactly why I need `a.a.a.Fx`

PranavPurwar:
You can add the blocks like how lightStatusBar block is added and does not use Fx for code generation

Edward-NewGate-KOC:
Now, everything’s working perfectly 🙂

PR #1400, opened 2024-11-20
fix: crash if blocks.json is invalid json

trindadedev13:


PranavPurwar:
A toast or notification would be better to let the user know its broken

trindadedev13:
> A toast or notification would be better to let the user know its broken

if the user open or create a palette, the json of blocks.xml is saved with valid json,  so I don't think it's necessary

PR #1443, opened 2024-11-28
Preview in Convert && More Widgets

trindadedev13:
[Example Video](https://jumpshare.com/s/37dzCRdv0KYm2uJwDcvP)

main issues:
You need to re-open fragment to update type.

also added MaterialSwitch preview, because i saw much people use it, I also plan to add preview for more views in future.

### DIFF
--- a/app/src/main/java/com/besome/sketch/editor/view/ViewPane.java
+++ b/app/src/main/java/com/besome/sketch/editor/view/ViewPane.java
@@ -424,122 +424,150 @@ public class ViewPane extends RelativeLayout {
             updateRelative(view, injectHandler);
         }
         if (classInfo.a("TextView")) {
-            TextView textView = (TextView) view;
-            updateTextView(textView, viewBean);
-            if (!classInfo.b("Button") && !classInfo.b("Switch")) {
-                textView.setGravity(viewBean.layout.gravity);
-            } else {
-                int gravity = viewBean.layout.gravity;
-                if (gravity == LayoutBean.GRAVITY_NONE) {
-                    textView.setGravity(Gravity.CENTER);
+            if (view instanceof TextView) {
+                TextView textView = (TextView) view;
+                updateTextView(textView, viewBean);
+                if (!classInfo.b("Button") && !classInfo.b("Switch")) {
+                    textView.setGravity(viewBean.layout.gravity);
                 } else {
-                    textView.setGravity(gravity);
+                    int gravity = viewBean.layout.gravity;
+                    if (gravity == LayoutBean.GRAVITY_NONE) {
+                        textView.setGravity(Gravity.CENTER);
+                    } else {
+                        textView.setGravity(gravity);
+                    }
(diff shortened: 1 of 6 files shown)


aikrq:
so many force pushes :skull:

trindadedev13:
> so many force pushes 💀

lol

hasrat-ali:
That's why local branch is good for testing.
Don't push untill it's almost ready

trindadedev13:
> That's why local branch is good for testing. Don't push untill it's almost ready

but for test, i need to push

trindadedev13:
most force-pushes are reverts of commits, or updating the branch with upstream

PranavPurwar:
What about opening PR *after* you've tested everything?

trindadedev13:
> What about opening PR _after_ you've tested everything?

Well, I opened the PR in case anyone who knows more about viewPane and wants to help me with the issue I mentioned

app/src/main/java/com/besome/sketch/editor/view/ViewPane.java
```diff
@@ -419,122 +435,150 @@ private void updateItemView(View view, ViewBean viewBean) {
...
-            }
-            if (classInfo.b("CircleImageView")) {
-                updateCircleImageView((ItemCircleImageView) view, injectHandler);
-            } else {
-                ((ImageView) view).setScaleType(ImageView.ScaleType.valueOf(viewBean.image.scaleType));
             }
         }
         if (classInfo.a("CompoundButton")) {
-            ((CompoundButton) view).setChecked(viewBean.checked != 0);
+            if (view instanceof CompoundButton) {
```
PranavPurwar, line 492: are these extra checks necessary? I think the views would already be of proper types

trindadedev13: well that fixed some  crashs related with cannot be cast

elfilibusterismo: Make it clear, crash that caused by changing the type of viewbean

elfilibusterismo:
> > What about opening PR _after_ you've tested everything?
> 
> Well, I opened the PR in case anyone who knows more about viewPane and wants to help me with the issue I mentioned

If you want to detect the changes of convert add another variable in ViewBean something like `preConvert` now remove the unnecessary check and do the checks in `ViewPane#g(ViewBean)`(not sure) , then add check where convert not equal preConvert

hasrat-ali:
Is there any use for this in reality?

elfilibusterismo:
> Is there any use for this in reality?

Yeah imo too, it doesn't really necessary now

elfilibusterismo:
> > Since we have `Direct XML editor` the layouts
> 
> Well if we think like this, Convert itself is no longer necessary

Not really, since Convert still allows custom ones. Your changes just seem a bit unnecessary now.

PR #1462, opened 2024-11-30
feat: add `edit`, `export`, and `import` actions for `customWidgets`

Edward-NewGate-KOC:


hasrat-ali:
When you want to push/merge yourself then why do you open the pull request?

Edward-NewGate-KOC:
> When you want to push/merge yourself then why do you open the pull request?

I’m trying to keep everything clean and double-check things before pushing, though.
I’m trying to be careful as much I can.

Do you think it’s better to avoid using pull requests?

hasrat-ali:
Can't you do that on your own machine? 

Is it really necessary to create another `merge` commit.
Anyway I just want `linear history` in this repo now.
Because now I know the pain of removing commits.

PR #1483, opened 2024-12-08
feat: add visibility property

trindadedev13:
https://t.me/sketchwarepro/223903/270070

Why draft?
i'll make another PR before finish this...


elfilibusterismo:
Is this really necessary? How would the user make changes to it when visibility is gone? By looking to the list of views or something? Isn't that a bad idea? What if I only have one widget and set visibility gone? How would I select it?

elfilibusterismo:
No more introducing another variable in ViewBean if it's only for adding a property for preview; use custom attributes (inject property) instead.

trindadedev13:
> No more introducing another variable in ViewBean if it's only for adding a property for preview; use custom attributes (inject property) instead.

bruh, users will have to use inject bruhhhhhhhh

elfilibusterismo:
> > No more introducing another variable in ViewBean if it's only for adding a property for preview; use custom attributes (inject property) instead.
> 
> bruh, users will have to use inject bruhhhhhhhh but okay, i"ll close again then 🥷

I literally just added a way for much easier for users to add an attributes in inject property

trindadedev13:
> > > No more introducing another variable in ViewBean if it's only for adding a property for preview; use custom attributes (inject property) instead.
> > 
> > 
> > bruh, users will have to use inject bruhhhhhhhh but okay, i"ll close again then 🥷
> 
> I literally just added a way for much easier for users to add an attributes in inject property

Yes

PR #1499, opened 2024-12-13
feat : Add Search Option In StringEditor

bestwall2:
I Have modified the adapter and the dialog of add string


PranavPurwar:
It would be nice if you can share screenshots of your PRs aswell

PR #1509, opened 2024-12-18
Create Translate to Portuguese

ErickHigino24:
I found a possible code to translate the app to Portuguese, I'm working to make a code that removes the translation and returns everything to English


ErickHigino24 (commented):
Why was it closed?

PR #1517, opened 2024-12-21
fix, perf: optimize Local Library Search

Edward-NewGate-KOC:
- Fixed ConcurrentModificationException caused by modifying the list while iterating
- The old method suffered from unnecessary threads, disk I/O, and adapter recreation, leading to delays and poor performance


trindadedev13 (approved):
lgtm

PR #1526, opened 2024-12-25
Redesign bottom layout of DesignActivity

elfilibusterismo:
Redesign project file selector and replace BuildingDialog
https://github.com/user-attachments/assets/2cb4c69a-931b-471e-9c2b-720254426b30


app/src/main/java/com/besome/sketch/design/DesignActivity.java
```diff
@@ -1284,10 +1281,8 @@ private void doInBackground() {
                     dialog.show();
                 });
             } catch (zy zy) {
-                isBuildFinished = true;
```
elfilibusterismo, line 1184: This should be included I think

app/src/main/java/com/besome/sketch/design/DesignActivity.java
```diff
@@ -1284,10 +1281,8 @@ private void doInBackground() {
...
                 activity.indicateCompileErrorOccurred(zy.getMessage());
             } catch (Throwable tr) {
-                isBuildFinished = true;
```
elfilibusterismo, line 1187: Same here

app/src/main/java/com/besome/sketch/design/DesignActivity.java
elfilibusterismo: @ilyassesalama

ilyassesalama: I believe those 2 lines weren't removed with the commit I just committed recently (?)

PR #1540, opened 2024-12-29
Project Resources Editor · Material 3 & Dynamic Colors Support · New Color Picker for resColors

Edward-NewGate-KOC:
- New Resources Editor
- Material3 and dynamic colors for projects
- New Color Picker for Material and project res colors

PR #1576, opened 2025-01-12
Fix ColorEditor & StringEditor ui

Bobur-777:


Edward-NewGate-KOC:
I don't think these changes are necessary because of #1540

I believe they will only cause merge conflicts 🥲

Bobur-777:
> I don't think these changes are necessary because of #1540
> 
> I believe they will only cause merge conflicts 🥲
> 
> 
Ohh i forgot about your new project resource editor 🙃👍

PR #1585, opened 2025-01-18
Update Hx.java

Mahmoud-Magd:
See this video to understand the problem 
https://t.me/sketchwarepro/203595/311674

Modified how Web client is added to the source code .. (Less code is generated)

And the original was adding all the events related to WebView .. Fixed that also

Now it always adds the webview.setWebViewClient(new WebViewClient()); And only adds the listener if the user added them 

Plus 

It doesn't add any other unwanted events

### DIFF
--- a/app/src/main/java/a/a/a/Hx.java
+++ b/app/src/main/java/a/a/a/Hx.java
@@ -348,11 +348,13 @@ public class Hx {
         private final String id;
         private final ArrayList<ComponentEvents> listeners = new ArrayList<>();
         private final boolean isViewBindingEnabled;
+        private final Gx classInfo; // to be used in generateEvent()
 
         private Event(Hx hx, String id, Gx classInfo, boolean isViewBindingEnabled) {
             this.hx = hx;
             this.id = id;
             this.isViewBindingEnabled = isViewBindingEnabled;
+            this.classInfo = classInfo; // to be used in generateEvent()
 
             String[] listeners = oq.b(classInfo);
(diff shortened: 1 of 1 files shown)


app/src/main/java/a/a/a/Hx.java
```diff
@@ -384,16 +385,34 @@ private Event(Hx hx, String id, Gx classInfo, boolean isViewBindingEnabled) {
 
         private String generateEvent() {
             StringBuilder sb = new StringBuilder(4096);
+            boolean webViewClientAdded = false; // Track if WebViewClient is added
+            
             for (ComponentEvents value : listeners) {
                 String event = value.generateEvent(id);
+                
+                // Ensure WebViewClient is always set
```
PranavPurwar, line 393: not the most efficient way im afraid

Mahmoud-Magd: Ok .. np ..

How u suggest to fix that bug?

hasrat-ali: i don't get any problem with WebView. so can you open an issue instead? if you cought any.

hasrat-ali (changes requested):

app/src/main/java/a/a/a/Hx.java
hasrat-ali: with details

hasrat-ali:
with

PR #1593, opened 2025-01-27
Fix list widget custom view in Direct XML Editor

elfilibusterismo:
This change fixes parsing in the XML editor for list widgets like RecyclerView by using the tools:listitem attribute it'll recognize listitem as the customView.


RoyTheReal:
Good job

PR #1605, opened 2025-03-01
Add a Chinese voice package and an Alibaba Cloud image to prevent some dependent libraries from being downloaded

blinlincom:
Add Chinese language


blinlincom:
@hasrat-ali Can it be merged?

hasrat-ali:
For repository: you should modify `/sdcard/.sketchware/libs/repositories.json` file.

For language: Use custom language `string.xml` file in `/sdcard/sketchware/localization/[string.xml]`

PranavPurwar:
I think we should either provide translations like that in this repository and use the localization file, when the language is not available. It's extremely difficult for users to find translations in the first place.

blinlincom:
> 我认为我们应该在这个存储库中提供这样的翻译，并在语言不可用时使用本地化文件。用户首先很难找到翻译。

Yes, I want to vote.

PR #1619, opened 2025-03-27
Added redirect to sketchware.pro

yoann256:
Added a redirect to sketchware.pro


trindadedev13:
hm, why any people will open https://sketchware-pro.github.io/Sketchware-Pro

yoann256:
maybe someone looking around the repo like me

PR #1622, opened 2025-03-29
fix: refine viewBinding implementation in block selector menu

Edward-NewGate-KOC:
- Changed the implementation of viewBinding so that `binding.` is no longer stored as part of the id when selecting a widget from the block selector menu. Instead, during source code generation, it checks each param type, and if it's a widget, it dynamically adds `binding.`
  
  - This update allows enabling or disabling the feature anytime without issues, unlike the previous approach, whether for `new or old projects`
  
  - Note: Widgets used in `ASD` will require manual adjustment

  - I've tested it with a big project from Sketchub, and it's working perfectly. Just need to double-check everything to make sure I didn’t miss anything

  - Also fixed a few bugs, like missing imports for some adapters and listeners

### DIFF
--- a/app/src/main/java/a/a/a/Fx.java
+++ b/app/src/main/java/a/a/a/Fx.java
@@ -717,7 +717,7 @@ public class Fx {
                 if (isViewBindingEnabled && paramAdapter.startsWith("binding.")) {
                     paramAdapter = paramAdapter.substring("binding.".length());
                 }
-                opcode = String.format("%s.setAdapter(new %s(%s));", param, Lx.a(paramAdapter), params.get(1));
+                opcode = String.format("%s.setAdapter(new %s(%s));", param, Lx.a(paramAdapter, isViewBindingEnabled), params.get(1));
                 break;
             case "listRefresh":
                 opcode = String.format("((BaseAdapter)%s.getAdapter()).notifyDataSetChanged();", params.get(0));
--- a/app/src/main/java/a/a/a/Fx.java
+++ b/app/src/main/java/a/a/a/Fx.java
@@ -8,18 +8,28 @@ import java.nio.CharBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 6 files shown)


app/src/main/java/a/a/a/Fx.java
```diff
@@ -8,18 +8,28 @@
...
 
     private static final Pattern PARAM_PATTERN = Pattern.compile("%m(?!\\.[\\w]+)");
     public final boolean isViewBindingEnabled;
+    private final ArrayList<String> viewParamsTypes = new ArrayList<>(List.of(
+            "%m.view", "%m.layout", "%m.textview", "%m.button", "%m.edittext", "%m.imageview", "%m.recyclerview",
+            "%m.listview", "%m.gridview", "%m.cardview", "%m.viewpager", "%m.webview", "%m.videoview", "%m.progressbar",
+            "%m.seekbar", "%m.switch", "%m.checkbox", "%m.spinner", "%m.tablayout", "%m.bottomnavigation", "%m.adview",
+            "%m.swiperefreshlayout", "%m.textinputlayout", "%m.ratingbar", "%m.datepicker", "%m.otpview", "%m.lottie",
+            "%m.badgeview", "%m.codeview", "%m.patternview", "%m.signinbutton", "%m.youtubeview"
+    ));
```
PranavPurwar, line 32: What about when users use custom view?

Edward-NewGate-KOC: you mean custom view for Alert dialogs, bottom sheets....???

Or just for adapters ??? ( It's working fine )

PranavPurwar: > you mean custom view for Alert dialogs, bottom sheets....???
> 
> Or just for adapters ??? ( It's working fine )

My bad, I was talking about the custom widget feature.

Edward-NewGate-KOC: Got it.
Anyway, the custom widgets won’t cause any problems; they are exactly like the built-in widgets

PR #1640, opened 2025-04-26
fix: Duplicate brackets when creating a HashMap

iKiwo:
Additionally, the "bean.id" was added so that when multiple HasMaps are created, they do not have the same variable.


trindadedev13:
I don't think changing this is really necessary, the brackets are exactly so that the variable name doesn't interfere with the others

PR #1672, opened 2025-06-13
Gradle plugin for Sketchware-Pro Build-In Libraries

Edward-NewGate-KOC:
- The idea is very simple instead of manually updating built-in libs, this feature makes updating them much easier — just by changing the version in build.gradle

- The plugin handles everything and puts the result directly into assets with better and more reliable handling of sub-dependencies automatically

- It also reduces the repo’s size since libs are no longer stored there

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -1,2 +1,3 @@
 /build
 /app/src/main/assets/libs/libs.zip
+/app/src/main/assets/libs/dexs.zip
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -1,1 +1,2 @@
 /build
+/app/src/main/assets/libs/libs.zip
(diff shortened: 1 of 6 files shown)


hasrat-ali:
@Edward-NewGate-KOC, you should look at [this](https://github.com/Sketchware-Pro/Library-Update-Manager) first. 😂 😂

Edward-NewGate-KOC:
> @Edward-NewGate-KOC, you should look at [this](https://github.com/Sketchware-Pro/Library-Update-Manager) first. 😂 😂

Damn, I remember you talked about this a long time ago on Discord, but I completely forgot it.

Anyway, I just checked the implementation. I didn't fully get how it works, but I think my base implementation is better — you can review it yourself to see

It handles sub-dependencies much more nicely (everything is managed automatically by Gradle, so we don't need to manually resolve incompatible or missing ones)

Also, why the hell are you generating a full BuiltInLibraries class? I think that's a bad idea

So in this PR, it instead generates a class with record for each library with here sub dependencies are neatly organized in record, and after linking it back to BuiltInLibraries, everything will work fine

Anyway, you can decide whether to close this PR, or use it, or merge it with yours in some form

PR #1681, opened 2025-06-21
Project Resources Editor · Material 3 & Dynamic Colors Support · New Color Picker for resColors

Edward-NewGate-KOC:
- New Resources Editor
 - Material3 and dynamic colors for projects
 - New Color Picker for Material and project res colors


RoyTheReal:
@Edward-NewGate-KOC The AndroidX Toolbar still isn't switched to the newer Material Toolbar in projects, also the background of the toolbar doesn't match what it's supposed to look like with material 3 + toolbar text is white instead of black

PR #1682, opened 2025-06-21
refactor, fix and optimize about activites and fragments

yoann256:
- AboutActivity.java got some fixes and optimizations
- BetaChangesFragment.java and ChangeLogFragment.java got fixes and optimization

### DIFF
--- a/app/src/main/java/pro/sketchware/activities/about/fragments/BetaChangesFragment.java
+++ b/app/src/main/java/pro/sketchware/activities/about/fragments/BetaChangesFragment.java
@@ -33,6 +33,8 @@ public class BetaChangesFragment extends Fragment {
     private int nextPage = 1;
     private boolean isLoading = false;
 
+    private static final int COMMITS_PER_PAGE = 100;
+
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
         binding = FragmentBetaChangesBinding.inflate(inflater, container, false);
@@ -59,7 +61,6 @@ public class BetaChangesFragment extends Fragment {
             @Override
             public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy) {
                 super.onScrolled(recyclerView, dx, dy);
-
                 if (!recyclerView.canScrollVertically(1) && !isLoading) {
                     adapter.showLoading();
                     loadCommitData(nextPage);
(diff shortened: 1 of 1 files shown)


app/src/main/java/pro/sketchware/activities/about/AboutActivity.java
```diff
@@ -42,7 +42,7 @@ public void onCreate(Bundle savedInstanceState) {
         binding = ActivityAboutAppBinding.inflate(getLayoutInflater());
         setContentView(binding.getRoot());
 
-        binding.toolbar.setNavigationOnClickListener(v -> finish());
```
PranavPurwar: Why?

yoann256: It gets overwritten by the initViews() fonction so it's useless

app/src/main/java/pro/sketchware/activities/about/AboutActivity.java
```diff
@@ -94,10 +94,10 @@ public void onPageSelected(int position) {
 
     private void initData() {
         network.get(Helper.getResString(R.string.link_about_team), response -> {
-            if (response != null) {
-                sharedPref.edit().putString("aboutData", response).apply();
-            } else {
+            if (response == null) {
```
trindadedev13: Why sir

yoann256: For me it what I normaly do in my code but if you'd like me to change it I can revert it

hasrat-ali:
After removing the useless changes, the changes that are left don't seem right to me.

And I also found a bug, it is not showing the current version, it is just showing new update.

But if you feel that we have made a mistake, please open a new pull request and explain the changes with full details.

Edward-NewGate-KOC:
> After removing the useless changes, the changes that are left don't seem right to me.
> 
> And I also found a bug, it is not showing the current version, it is just showing new update.
> 
> But if you feel that we have made a mistake, please open a new pull request and explain the changes with full details.

It's not his bug. The issue is simply because the commit hash doesn't exist in the main branch, which is why it appears that way
If you push his changes to the main branch, it will work fine

Btw, I reviewed his changes, and I believe they're unnecessary anyway

Edward-NewGate-KOC:
Oh, you reverted some of his changes. I just noticed that

I think you should merge it, as his changes will prevent the app from crashing if GitHub API temporarily blocks the user's IP due to too many requests

PR #1692, opened 2025-06-25
fix: Display correct options for each activities or fragments in AddViewActivity.java

yoann256:
It would be a great thing that types other than activities has manifest settings disabled.
What I mean is that Fragments, DialogFragments and BottomSheetDialogFragments does not need to be declared in AndroidManifest.xml, and as such, don't need any options.
This PR will be effectively what I say, as it disables manifest options for such types.
Still WIP.


trindadedev13:
well, I don't think it's a good idea to just hide the views in a dry way, a transition anim would be nice

yoann256:
> well, I don't think it's a good idea to just hide the views in a dry way, a transition anim would be nice



> well, I don't think it's a good idea to just hide the views in a dry way, a transition anim would be nice

Okay I'll work on that

yoann256:
> well, I don't think it's a good idea to just hide the views in a dry way, a transition anim would be nice

What do you think about this?

https://github.com/user-attachments/assets/5ab33d8b-b7a2-4f8b-aa7e-c0fbbfe079ab

trindadedev13:
> > well, I don't think it's a good idea to just hide the views in a dry way, a transition anim would be nice
> 
> What do you think about this?
> 
>  Screenrecorder-2025-06-26-11-08-12-645.mp4

lgtm

yoann256:
I think instead of disabling manifest options when editing I'll just stop being able to edit it at all

trindadedev13:
> I think instead of disabling manifest options when editing I'll just stop being able to edit it at all

what

yoann256:
> > I think instead of disabling manifest options when editing I'll just stop being able to edit it at all
> 
> what

what i mean is that when you edit an activity from the layout list you can only change orientation and keyboard settings (manifest options) and the name of the activity, fragment... you can't change so that leaves us with nothing to change so instead i'll just remove the way to edit the activity from the layout list

yoann256:
> > > I think instead of disabling manifest options when editing I'll just stop being able to edit it at all
> > 
> > 
> > what
> 
> what i mean is that when you edit an activity from the layout list you can only change orientation and keyboard settings (manifest options) and the name of the activity, fragment... you can't change so that leaves us with nothing to change so instead i'll just remove the way to edit the activity from the layout list

i ended up not doing that cuz lazy

yoann256:
pretty sure it's ready

trindadedev13:
Thank You!

PR #1735, opened 2025-07-05
feat: exit app completely

kkgit2008:
Now when we click back twice,the app will not run in background.


hasrat-ali:
Pls don't spam with unnecessary pull requests.

PR #1738, opened 2025-07-05
Update in MainActivity

AnBui2004:
- Appbar changes background color to match Fragment background.
- Change icons in dialogs in MainActivity.
- org.gradle.jvmargs=-Xmx2048m -> org.gradle.jvmargs=-Xmx4096m.


AnBui2004:
image
image
image

app/src/main/res/layout/main.xml
```diff
@@ -15,10 +15,10 @@
             android:id="@+id/appbar"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:background="@android:color/transparent"
             app:elevation="0dp"
-            app:liftOnScroll="false"
-            app:liftOnScrollColor="@android:color/transparent">
+            app:liftOnScroll="false">
+<!--            android:background="@android:color/transparent"-->
```
TheRealAshik: Why not permanently removed 😴

StudioAsInc (approved):
FINE !!!

app/src/main/res/layout/main.xml
AnBui2004: Because when you open the app it doesn't have a different color than the bottom. Haven't you tried it?

TheRealAshik: I have tried! I WAS TALKING ABOUT instead of commenting on the code. You could have removed it 🛠

AnBui2004: I just wanted to keep them so the person who wrote them could see how they were changed more easily.

gradle.properties
yoann256, line 1: why

AnBui2004: I built apk on Android Studio and it's not enough. For projects like Sketchware Pro it deserves more.

yoann256: ok

gradle.properties
```diff
@@ -6,7 +6,7 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8 -XX:+UseParallelGC -Dorg.gradle.warning.mode=all
+org.gradle.jvmargs=-Xmx4096m -Dfile.encoding=UTF-8 -XX:+UseParallelGC -Dorg.gradle.warning.mode=all
```
hasrat-ali: Not everyone has that much free RAM, so it would be better to keep it as it is.

hasrat-ali (changes requested):

app/src/main/java/pro/sketchware/activities/main/activities/MainActivity.java
```diff
@@ -264,12 +266,17 @@ public void onFinish() {
         }
 
         binding.bottomNav.setOnItemSelectedListener(item -> {
+            AppBarLayout _appbar = findViewById(R.id.appbar);
```
hasrat-ali: why don't you use ViewBinding here?

AnBui2004: I will update.
image

gradle.properties
AnBui2004: OK, so I will set it as before each request.

TheRealAshik: > Not everyone has that much free RAM, so it would be better to keep it as it is.

Can't it be **Dynamicly change** according to the device specification or adjustable in settings in some way? or trick?

AnBui2004:
Set the heap memory for the JVM according to the total RAM the device has. (Raw code)

I haven't found a place to start building APK or AAB to put a more appropriate number.
image

TheRealAshik:
> Set the heap memory for the JVM according to the total RAM the device has. (Raw code)
> 
> I haven't found a place to start building APK or AAB to put a more appropriate number.
> image

Where did you get this? Setting full memory uses will cause problems & lags. 50-70% looks better deal. [2-4gb] anything btwn this is good

AnBui2004:
> > Set the heap memory for the JVM according to the total RAM the device has. (Raw code)
> > I haven't found a place to start building APK or AAB to put a more appropriate number.
> > image
> 
> Where did you get this? Setting full memory uses will cause problems & lags. 50-70% looks better deal. [2-4gb] anything btwn this is good

That said, I haven't found a place where Sketchware Pro starts building APKs or AABs so this is just a suggestion. If I find out, everything will be easier.

TheRealAshik:
> That said, I haven't found a place where Sketchware Pro starts building APKs or AABs, so this is just a suggestion. If I find out, everything will be easier.


Dynamically changable according to device will be best implementation if possible

AnBui2004:
> > That said, I haven't found a place where Sketchware Pro starts building APKs or AABs, so this is just a suggestion. If I find out, everything will be easier.
> 
> Dynamically changable according to device will be best implementation if possible

My idea is that when the user taps the build APK or AAB button, it will get the current free RAM and from there it can set a suitable number and then start building the APK or AAB.

TheRealAshik:
> My idea is that when the user taps the build APK or AAB button, it will get the current free RAM and from there it can set a suitable number and then start building the APK or AAB.


Is it possible to do this programmatically? I mean `gradle.properties`

AnBui2004:
> > My idea is that when the user taps the build APK or AAB button, it will get the current free RAM and from there it can set a suitable number and then start building the APK or AAB.
> 
> Is it possible to do this programmatically? I mean `gradle.properties`

My idea is that every time it builds an APK or AAB it checks and sets, and this happens in all projects. But with the current code it's always fixed.

AnBui2004:
Automatically set the heap memory for JVM according to the free RAM on the device. It will use 57% of the free RAM at each APK or AAB build.

image
image

TheRealAshik:
> It will use 57% of the free RAM at each APK or AAB build.

BTW why specifically 57%

AnBui2004:
> > It will use 57% of the free RAM at each APK or AAB build.
> 
> BTW why specifically 57%

Do you want to increase it because this is a safe ratio.

AnBui2004:
I saw someone change the background color so I removed the appbar color change.

AnBui2004:
Now it is using 66% of total free RAM.

AnBui2004:
It's still 512MB even though gradle.properties has changed.

TheRealAshik:
> It's still 512MB even though gradle.properties has changed.

What is **512mib**? If you are talking about allocated RAM. your code looks ok!

AnBui2004:
> > It's still 512MB even though gradle.properties has changed.
> 
> What is **512mib**? If you are talking about allocated RAM. your code looks ok!

But Sketchware doesn't use the set memory, it still uses 512MB and when using R8 with quite complex projects it runs out of ram.

aikrq:
Sketchware doesn't use Gradle for building projects.

PR #1742, opened 2025-07-07
feat: Add android:tint property support for ImageView

alejandroHAR:
Adds support for the `android:tint` property in ImageView, allowing users to apply a color filter to images directly from the visual editor's property panel.


alejandroHAR:
image
image
image
image

alejandroHAR:
image
image
image
image

TheRealAshik:
Genius idea 💡

app/src/main/java/a/a/a/Ox.java
```diff
@@ -585,6 +586,21 @@ private void writeImageScaleType(XmlBuilder nx, ViewBean viewBean) {
         }
     }
 
+    private void writeImageTint(XmlBuilder nx, ViewBean viewBean) {
+        var injectHandler = new InjectAttributeHandler(viewBean);
+        Set<String> toNotAdd = readAttributesToReplace(viewBean);
+
+        if (viewBean.image.tintColor != 0xffffff) {
+            if (!toNotAdd.contains("android:tint") && !injectHandler.contains("tint")) {
```
Edward-NewGate-KOC, line 594: it's better to use `app:tint` instead

alejandroHAR: Yes, I remembered that after sending the PR 😂 There are no backwards compatibility issues, right?

Edward-NewGate-KOC: exactly, that's why you should use `app:tint`

TheRealAshik: So supportive 💯

app/src/main/java/a/a/a/Ox.java
```diff
@@ -585,6 +586,21 @@ private void writeImageScaleType(XmlBuilder nx, ViewBean viewBean) {
...
+                if (viewBean.image.resTintColor != null && !viewBean.image.resTintColor.isEmpty()) {
+                    nx.addAttribute("android", "tint", "@color/" + viewBean.image.resTintColor);
```
Edward-NewGate-KOC, line 596: Have you tested picking colors from Material palette or colors.xml???

I believe this might generate incorrect code like:

- `android:tint="@color/@color/colorPrimary"`
- `android:tint="@color/?colorPrimary"`

So, you can fix it by removing ` "@color" +` Just pass the color reference directly from the color picker

Check this behavior I'm not sure about it, cause I can't test it myself right now

app/src/main/java/com/besome/sketch/beans/ImageBean.java
```diff
@@ -34,16 +34,24 @@ public ImageBean[] newArray(int size) {
     public int rotate;
     @Expose
     public String scaleType;
+    @Expose
+    public int tintColor;
+    @Expose
+    public String resTintColor;
```
hasrat-ali, line 40: i think, you should not add new keys to Bean claases. you need to find another way.

hasrat-ali (changes requested):

elfilibusterismo:
@hasrat-ali is right, I already added a new way using custom attributes(inject) property 
You can see the implementation for elevation here:
https://github.com/Sketchware-Pro/Sketchware-Pro/blob/main/app%2Fsrc%2Fmain%2Fjava%2Fcom%2Fbesome%2Fsketch%2Feditor%2Fproperty%2FPropertyInputItem.java#L564-L612

And here’s how it's applied to the view:
https://github.com/Sketchware-Pro/Sketchware-Pro/blob/main/app%2Fsrc%2Fmain%2Fjava%2Fcom%2Fbesome%2Fsketch%2Feditor%2Fview%2FViewPane.java#L568-L570

alejandroHAR:
> @hasrat-ali is right, I already added a new way using custom attributes(inject) property You can see the implementation for elevation here: https://github.com/Sketchware-Pro/Sketchware-Pro/blob/main/app%2Fsrc%2Fmain%2Fjava%2Fcom%2Fbesome%2Fsketch%2Feditor%2Fproperty%2FPropertyInputItem.java#L564-L612
> 
> And here’s how it's applied to the view: https://github.com/Sketchware-Pro/Sketchware-Pro/blob/main/app%2Fsrc%2Fmain%2Fjava%2Fcom%2Fbesome%2Fsketch%2Feditor%2Fview%2FViewPane.java#L568-L570

I don't understand the comment. This is for the "android:tint" in the ImageView properties with a color picker. And for now, you can't select colors from the Injector. The color picker used for text, background, etc. colors is used instead.

app/src/main/java/com/besome/sketch/beans/ImageBean.java
alejandroHAR: why?

PR #1756, opened 2025-07-10
feat: enhance library downloader with advanced progress tracking and improved UX

alejandroHAR:
I redesigned this update from the ground up; the previous one didn't work as expected. I also added a history of previously downloaded libraries.

### DIFF
--- a/app/src/main/java/dev/aldi/sayuti/editor/manage/DependencyDownloadAdapter.java
+++ b/app/src/main/java/dev/aldi/sayuti/editor/manage/DependencyDownloadAdapter.java
@@ -1,0 +1,102 @@
+package dev.aldi.sayuti.editor.manage;
+
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import pro.sketchware.databinding.ItemDependencyDownloadBinding;
+
+
+
+public class DependencyDownloadAdapter extends RecyclerView.Adapter<DependencyDownloadAdapter.ViewHolder> {
+
+    private final List<DependencyDownloadItem> dependencies = new ArrayList<>();
+
+    @NonNull
+    @Override
+    public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+        ItemDependencyDownloadBinding binding = ItemDependencyDownloadBinding.inflate(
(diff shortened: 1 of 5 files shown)


aikrq:
Why don't you follow the contribution guidelines?

PranavPurwar:
@alejandroHAR Can you attach some screenshots of the changes along with the built APK file?

alejandroHAR:
> @alejandroHAR Can you attach some screenshots of the changes along with the built APK file?

Yes, let me make some changes and I'll send it.

PranavPurwar:
@alejandroHAR I'm waiting...

alejandroHAR:
> @alejandroHAR I'm waiting...

PranavPurwar:
I think you should remove the suggestions from the input and why is there split button? does it show any new options on clicking? and when download for some library finishes change the progress bar with a check icon. I think that will improve the experience

alejandroHAR:
@PranavPurwar Input suggestions are initially empty, and are saved as the user adds dependencies. Sometimes we delete dependencies instead of disabling them and we have to search the repo again.

PranavPurwar:
> @PranavPurwar Input suggestions are initially empty, and are saved as the user adds dependencies. Sometimes we delete dependencies instead of disabling them and we have to search the repo again.

I believe most people just disable the libraries rather than deleting. Moreover if the library is already downloaded, isn't it counterintuitive to still show it to the user?

TheRealAshik:
> > @PranavPurwar Input suggestions are initially empty, and are saved as the user adds dependencies. Sometimes we delete dependencies instead of disabling them and we have to search the repo again.
> 
> I believe most people just disable the libraries rather than deleting. Moreover if the library is already downloaded, isn't it counterintuitive to still show it to the user? 

Simply check if the user has it or not?

TheRealAshik:
Feat request: Add an update option that will  show another latest version of that dependency is available!!!

alejandroHAR:
> Feat request: Add an update option that will  show another latest version of that dependency is available!!!

I was thinking exactly the same thing. But for that, we'd have to change other things, maybe in another PR.

alejandroHAR:
> > @PranavPurwar Input suggestions are initially empty, and are saved as the user adds dependencies. Sometimes we delete dependencies instead of disabling them and we have to search the repo again.
> 
> I believe most people just disable the libraries rather than deleting. Moreover if the library is already downloaded, isn't it counterintuitive to still show it to the user?

is ready

TheRealAshik:
> is ready

Finally! 😵‍💫🔥

TheRealAshik:
Do I have to die to see this feat?

PR #1758, opened 2025-07-10
fix: check firebase init before calling firebase messaging service

waelchateur:
this will prevent app crash in case firebase isnt configured  or didnt initialised at startup.

ps : the crash is happening Already for forks cause contributors doesnt have access to hiden files aka secrets


waelchateur:
now pr should be fixed 😅, i didnt test this cause the app will crash

waelchateur:
now pr should be fixed 😅, it was a blind Pr i didnt test

PR #1759, opened 2025-07-10
fix: check firebase init before calling firebase messaging service

waelchateur:
this will prevent app crash in case firebase isnt configured or didnt initialised at startup for project store

ps : the crash is happening Already for forks contributors doesnt have access to hiden files aka secrets

the pr is working now 👌


trindadedev13:
you could have just updated the other PR...

waelchateur:
> you could have just updated the other PR...

i am using phone somehow reverting and forcepushin after adding the missing statement in onpostcreat did not includ the old changes in the same commit so i created new pr

PranavPurwar (approved):

PR #1779, opened 2025-07-21
Fix creating null xml drawables

Bobur-777:


app/src/main/java/com/besome/sketch/design/DesignActivity.java
```diff
@@ -768,7 +768,8 @@ private void askIfToRestoreOldUnsavedProjectData() {
                     jC.a(sc_id).a(jC.b(sc_id));
                 }
                 if (q) {
-                    jC.a(sc_id).b(jC.d(sc_id));
+                    // This method is replaces not exist images to default_image, I removed it because it changes vector images to default_images
```
Edward-NewGate-KOC: You can simply delete the line and explain the reason in the commit message

Edward-NewGate-KOC (changes requested):

app/src/main/java/com/besome/sketch/design/DesignActivity.java
Bobur-777: I am not removed because maybe it is useful for later

PR #1795, opened 2025-07-26
fix: add missing return statement in fragment adapter

waelchateur:
this will fix compiling errors when using  the FragmentAdapter component

### DIFF
--- a/app/src/main/java/mod/agus/jcoderz/editor/event/ManageEvent.java
+++ b/app/src/main/java/mod/agus/jcoderz/editor/event/ManageEvent.java
@@ -6,6 +6,7 @@ import a.a.a.Gx;
 import a.a.a.Lx;
 import mod.hilal.saif.events.EventsHandler;
 import pro.sketchware.R;
+import androidx.fragment.app.Fragment;
 
 public class ManageEvent {
 
@@ -554,10 +555,10 @@ public class ManageEvent {
                     "}";
             case "onFragmentAdded" -> //noinspection DuplicateExpressions
                     "@Override\r\n" +
-                            "public Fragment getItem(int _position) {\r\n" +
-                            (!eventLogic.isEmpty() ? eventLogic + "\r\n" :
-                                    "return null;\r\n") +
-                            "}";
+                           "public Fragment getItem(int _position) {\r\n" +
+                           (!eventLogic.isEmpty() ? eventLogic + "\r\n" : "") +
+                                    "return new Fragment();\r\n" +
+                           "}";
             case "onTimeSet" -> "@Override\r\n" +
                     "public void onTimeSet(TimePicker _timePicker, int _hour, int _minute) {\r\n" +
                     eventLogic + "\r\n" +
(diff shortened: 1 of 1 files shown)


app/src/main/java/mod/agus/jcoderz/editor/event/ManageEvent.java
```diff
@@ -554,10 +555,10 @@ public static String f(String targetId, String eventName, String eventLogic) {
                     "}";
             case "onFragmentAdded" -> //noinspection DuplicateExpressions
                     "@Override\r\n" +
-                            "public Fragment getItem(int _position) {\r\n" +
-                            (!eventLogic.isEmpty() ? eventLogic + "\r\n" :
-                                    "return null;\r\n") +
-                            "}";
+                           "public Fragment getItem(int _position) {\r\n" +
+                           (!eventLogic.isEmpty() ? eventLogic + "\r\n" : "") +
+                                    "return new Fragment();\r\n" +
```
aikrq, line 560: why?

waelchateur: because it should always return a fragment but recently this was edited and return is removed & causing projects errors when using the fragmentadapter component 

for example : these are errors showing for every projects using tablayout
Screenshot_20250726_020335_Sketchware Pro

waelchateur: @aikrq if u r asking why it is returning empty fragment instead of null like what it was before ? okey i can tell that its safer and won't cause null pointer exceptions, more than this the app doesn't crash from missing logic

waelchateur: & same for the ontabadded it should return but its not and always project will throw errors

app/src/main/java/mod/agus/jcoderz/editor/event/ManageEvent.java
```diff
@@ -6,6 +6,7 @@
 import a.a.a.Lx;
 import mod.hilal.saif.events.EventsHandler;
 import pro.sketchware.R;
+import androidx.fragment.app.Fragment;
```
Edward-NewGate-KOC, line 9: We don't need this import, you just added `new Fragment()` in string

app/src/main/java/mod/agus/jcoderz/editor/event/ManageEvent.java
Edward-NewGate-KOC, line 560: I understand your idea, you're trying to avoid the same error that occurred in old projects when Hasrat removed `return null;`, but now you're bringing that logic back in a different form

Previously, we asked users to manually edit the event to fix that error. If we merge this changes, it could reintroduce a new error in projects using the nightly builds

In short: this change might fix one thing, but it breaks another, So i think you should revert all the changes except replacing `null` with `new Fragment()` to avoid the exception

Edward-NewGate-KOC (changes requested):

app/src/main/java/mod/agus/jcoderz/editor/event/ManageEvent.java
waelchateur: my idea was to prevent projects from crash if the adapter logic is empty and user still working on it and need to test so an empty fragment will be returned instead, but now i guess better i close this PR since its not needed anymore and it was already handled before and to prevent breaking projects that already using the idea that hasrat describe.

Edward-NewGate-KOC:
I think the best way to fix this issue is to use JavaParser to check for missing return statements.

waelchateur:
> I think the best way to fix this issue is to use JavaParser to check for missing return statements.

i have closed the Pr i will leave it to you and the others for better dicisions , and thank you so mutch for the review and your time

PR #1832, opened 2025-08-08
fix: string.xml button in input dialog not working correcly

Bobur-777:
> Fixed text dialog's neutral button clickable


trindadedev13:
Thank you!

PR #1833, opened 2025-08-08
added new blocks for rest api

FabioSilva11:
dedicated blocks for rest api


PranavPurwar:
No new blocks will be added directly to Sketchware Pro. Please use the custom blocks feature to create and share blocks.

PR #1868, opened 2025-08-30
Fixed pressing the run button while canceling it would start building while canceling.

annbeep62:
`fix: Fixed pressing the run button while canceling it would start building while canceling.`


trindadedev13:
Can you show when this issue happens? I tried it here and run button is working properly.

annbeep62:
> Can you show when this issue happens? I tried it here and run button is working properly.

I don't have a video because I'm using a custom version and have fixed it.

When you click Stop, `if (currentBuildTask != null && !currentBuildTask.canceled && !currentBuildTask.isBuildFinished)` is still true and it will start canceling. But while it is canceling and you click Stop, the condition in that line of code is no longer `true` because on the previous stop click `canceled` was immediately set to `true` in `public void cancelBuild()`. So `!currentBuildTask.canceled` is `false` and it is no longer prevented from starting the build by `return;`. And now it is canceling and building at the same time.

trindadedev13:
https://github.com/user-attachments/assets/d0d5649b-0d5f-47e9-8547-a850956779e8

annbeep62:
> Screen_Recording_20250830_141243_Android.System.mp4

It will cause noticeable lag for quite complex projects because there is more than one build going on. And your project is empty so you don't notice.

AnBui2004:
https://files.catbox.moe/jlnme8.mp4

As you can see in this video I kept pressing the Stop button and it kept getting laggy as more and more builds were happening. Since this is a fairly complex project, it doesn't have time to cancel the build process before clicking the Stop button while it's canceling.

hasrat-ali:
Just disable the run button. 
Isn't that simple?

hasrat-ali:
Closing this PR in favor of a UI-driven approach.

Instead of silently ignoring clicks during the cancellation phase, we have implemented a change to explicitly disable the "Run/Stop" button when the build is being canceled. This provides better feedback to the user that the system is busy stopping the process and prevents concurrent build attempts.

Thank you for identifying the issue!

